PR #14, opened 2017-10-25
Added an activate button example

sebasguts:
@nthiery Could you have a look at this example? I am not sure what (or if anything) should go into thebelab, or if it should just stay in the example directory. Non of those features seem very specific, but I think Nicolas wanted to have some standartized activate button.

### DIFF
--- a/examples/activate_button.html
+++ b/examples/activate_button.html
@@ -1,0 +1,54 @@
+<html>
+<head>
+  <meta HTTP-EQUIV="Content-Type" CONTENT="text/html; charset=UTF-8">
+  <title>Thebe activate button</title>
+  <link rel="stylesheet" type="text/css" href="index.css" />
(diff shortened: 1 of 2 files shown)


nthiery:
I finally get to check this out. Thanks Sebastian!

I agree, there is nothing very complicated here. Still one good page of code to write/copy, and I'd expect most users of ThebeLab not to be fluent with javascript. Given that the feature would be already used in at least two applications (Sage's and GAP's doc), I would tend to merge it in. But I see the point of resisting feature creep. I will let @minrk decide what's the best route between:
- making it a standard feature
- keeping it as an example for now; and if the feature gets used on a regular basis, make it standard.

minrk:
Thanks! Let's keep it as a small example for now, and we can look into merging it in at some point in the future.

PR #19, opened 2017-11-13
Added function that splits at output promt.

sebasguts:
I don't know if this could be sensibly merged with the splitCell function and keeping both functions as easy as possible.


nthiery:
This looks good. Could you add an example of usage in examples/ ? To serve as a test too.

PR #31, opened 2018-01-25
Make activate button example automatically request kernel

sebasguts:


nthiery:
The new behavior is more natural, and it's a tiny change. Let's go for it.

PR #39, opened 2018-01-28
Added the possibility to add predefined ouput to cells

sebasguts:
via the data-output attribute. Rendering the output can be
controlled by the defaultOutput option

### DIFF
--- a/examples/predefined_output.html
+++ b/examples/predefined_output.html
@@ -1,0 +1,42 @@
+<html>
+<head>
+  <meta HTTP-EQUIV="Content-Type" CONTENT="text/html; charset=UTF-8">
+  <title>Thebe Lab examples</title>
+  <link rel="stylesheet" type="text/css" href="index.css" />
+
+  <!-- Configure and load Thebe !-->
+  <script type="text/x-thebe-config">
+    {
+      bootstrap: true,
+      binderOptions: {
+        repo: "binder-examples/requirements",
+      },
+      kernelOptions: {
+        name: "python3",
+      },
+      defaultOutput: true
+    }
+  </script>
+<!--   <script type="text/javascript" src="https://unpkg.com/thebelab@^0.1.0"></script> -->
+  <!-- or to use a local build: -->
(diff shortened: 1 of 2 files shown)


nthiery:
Hi @sebasguts,
I am not sure what's the exact use case you have in mind. Is the point that, currently, upon starting thebelab all outputs are cleared, whereas you would like to keep display the original value (if existing), until it's replaced by the result of the calculation?

sebasguts:
Yep, you are right.

Currently, there is no possibility to keep the output from example cells from manuals, or add a default output. This PR enables you to equip cells with a default output, that is placed as it was output from the kernel. This output gets replaced when you hit run or shift+enter.

nthiery:
On Sun, Jan 28, 2018 at 06:23:02AM -0800, Sebastian Gutsche wrote:
>    Currently, there is no possibility to keep the output from example
>    cells from manuals, or add a default output. This PR enables you to
>    equip cells with a default output, that is placed as it was output from
>    the kernel. This output gets replaced when you hit run or shift+enter.

Ok, that's a nice feature indeed.

Would it be possible to change the syntax to something like:

        <pre data-executable="true" data-language="python"">print("Hello!")</pre>
        <pre data-output="true">Hello!</pre>

This would be somewhat more natural, and in particular not require
painful quoting for long or complex outputs.

The downside is that it's probably a bit more complicated to implement
in Thebe, requiring to mess not just with on node in the Dom, but too
consecutive nodes.

Cheers,

sebasguts:
If you use the `stripPrompts` command to render the cells, everything stays the same as it is, as it removes the output and puts in the `data-output` attribute. But that is also a nice option. Will think about how this can be done.

akhmerov:
I have a question about the current implementation: would it work if the output isn't text but rather a general html? (think bokeh or plotly plot for example)

akhmerov:
I believe this was now addressed in #105.

sebasguts:
Definitely, thanks :)

PR #51, opened 2018-06-05
Fix link to online demo in examples/README.md

slel:
Change link target from [0] which was "404 Not Found" to [1].

- [0] https://minrk.github.io/thebelab/examples/
- [1] https://minrk.github.io/thebelab/demo


nthiery:
Thanks Sam for catching this! I fixed it directly on master (pointing to index rather demo as it was meant for).

PR #64, opened 2018-07-29
Add custom widget example

SylvainCorlay:
I am also cleaning up the loading of custom widgets.

maps

### DIFF
--- a/examples/index.html
+++ b/examples/index.html
@@ -68,6 +68,7 @@ plt.plot(x, np.cos(x))</pre>
   <h2>More examples</h2>
 
   <ul>
+  <li><a href="widgets.html">Making use of Jupyter interactive widgets</a>;
   <li><a href="status_field.html">Status field and styling</a>;
   <li><a href="activate_button.html">Activate/Status button</a>;
   <li><a href="prompts.html">Alternative computational environments; code cells with prompts and outputs</a>;
--- a/src/thebelab.js
+++ b/src/thebelab.js
@@ -22,11 +22,23 @@ import {
 import {
   ThebeManager
 } from './manager';
+import {
+  requireLoader
+} from './loader';
+
 import { Mode } from "@jupyterlab/codemirror";
 
 import "@jupyterlab/theme-light-extension/static/index.css";
 import "@jupyter-widgets/controls/css/widgets.built.css";
 import "./index.css";
+
+import * as base from '@jupyter-widgets/base';
+import * as controls from '@jupyter-widgets/controls';
+
+if (typeof window !== "undefined" && window.define !== "undefined") {
+    window.define("@jupyter-widgets/base", base);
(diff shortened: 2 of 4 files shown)


examples/index.html
```diff
@@ -4,21 +4,23 @@
...
 
+  <script src="https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.4/require.min.js">
+  </script>
+
   <!-- Configure and load Thebe !-->
   <script type="text/x-thebe-config">
     {
       bootstrap: true,
       binderOptions: {
-        repo: "binder-examples/requirements",
```
SylvainCorlay: TODO: undo this change. ipyleaflet is needed for that example. It should probably be added to `binder-examples/requirements`, or we could make a new `binder-examples/thebelab-demo`.

examples/index.html
```diff
@@ -4,21 +4,23 @@
   <title>Thebe Lab examples</title>
   <link rel="stylesheet" type="text/css" href="index.css" />
 
+  <script src="https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.4/require.min.js">
+  </script>
```
SylvainCorlay: requirejs is required for loading custom widgets on unpkg.

webpack.config.js
```diff
@@ -97,4 +92,5 @@ module.exports = {
       },
     ],
   },
+  externals: ["'@jupyter-widgets/base'", "'@jupyter-widgets/controls'"]
```
SylvainCorlay: > So this is really weird. This addition of an external statement causes uglifyjs to complain.
>
> ERROR in index.js from UglifyJs
> Unexpected character '@' [external "@jupyter-widgets/base":1,0][index.js:26225,17]
> When disabling uglify, the webpack stage completes properly, but at runtime I get invalid JS syntax in the generated bundle, namely:
> 
> module.exports = @jupyter-widgets/base;
> which causes the browser to complain: Uncaught SyntaxError: Invalid or unexpected token.
>
> I have tried to understand what is going on there and could not find the issue,

Adding a second set of quotes seems to fix it but I don't know what is going on here.

src/thebelab.js
```diff
@@ -22,12 +22,24 @@ import {
...
 import { Mode } from "@jupyterlab/codemirror";
 
 import "@jupyterlab/theme-light-extension/static/index.css";
 import "@jupyter-widgets/controls/css/widgets.built.css";
 import "./index.css";
 
+import * as base from '@jupyter-widgets/base';
+import * as controls from '@jupyter-widgets/controls';
+
+if (typeof window !== "undefined" && window.require !== "undefined") {
```
minrk: check for window.define instead?

SylvainCorlay:
@minrk I think that this is ready to go. It requires publishing `thebelab` `0.2.1` to npmjs.

S.

PR #66, opened 2018-08-01
Use fixed ref for ipyleaflet binder

SylvainCorlay:


SylvainCorlay:
Cool, that should prevent many binder builds :)

PR #70, opened 2018-08-03
Add restart button

### LINKED ISSUE #69: Means to restart kernel
I am testing thebelab with some C++ kernels for which restarts are more frequent, and that would be a great usability improvements to have a means to restart kernels.

SylvainCorlay:
Fixes #69 

Seems to work fine. (Adding a restart button next to the run.)

### DIFF
--- a/src/thebelab.js
+++ b/src/thebelab.js
@@ -167,6 +167,11 @@ function renderCell(element, options) {
       .text("run")
       .click(execute)
   );
+  $cell.append(
+    $("<button class='thebelab-button thebelab-restart-button'>")
+      .text("restart")
+      .click(restart)
+  );
   let kernelResolve, kernelReject;
   let kernelPromise = new Promise((resolve, reject) => {
     kernelResolve = resolve;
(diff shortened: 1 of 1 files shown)


SylvainCorlay:
ping @minrk.

adds a `restart` button next to the `run`.

SylvainCorlay:
@minrk when this gets in, I will add a C++ example!

SylvainCorlay:
up?

src/thebelab.js
```diff
@@ -167,6 +167,11 @@ function renderCell(element, options) {
       .text("run")
       .click(execute)
   );
+  $cell.append(
+    $("<button class='thebelab-button thebelab-restart-button'>")
+      .text("restart")
```
minrk: Can you add `.attr('title', 'restart the kernel')` so there's info about what it does? Maybe a similar 'run this cell' title on the run button while we're at it.

SylvainCorlay: Done.

src/thebelab.js
```diff
@@ -164,9 +164,14 @@ function renderCell(element, options) {
   $cell.append($cm_element);
   $cell.append(
     $("<button class='thebelab-button thebelab-run-button'>")
-      .text("run")
+      .text("run this cell")
       .click(execute)
   );
+  $cell.append(
+    $("<button class='thebelab-button thebelab-restart-button'>")
+      .text("restart the kernel")
```
minrk: Sorry, not `.text`, which changes the button label, but `.attr("title", "restart the kernel")` to give hover text. Leave the button text as "run" and "restart"

PR #78, opened 2018-08-20
Adds support for gitlab and git repositories

Miniland1333:
Added a new optional binderOptions attribute "repoProvider" to specify an alternative provider such as gitlab or git.

Github is still the assumed default and this change does not affect existing installations.


Miniland1333:
@moorepants This is the commit that I'm waiting on for Thebelab

moorepants:
@Miniland1333 can you make use of your fork to run things on Libretexts until it is decided if this is merged or not?

minrk:
Great, thanks!

PR #98, opened 2019-01-11
Added option for codeMirrorConfig

### LINKED ISSUE #97: Expose how to pass extra configuration to codemirror
It seems that codemirror is quite configurable (e.g. https://codemirror.net/demo/theme.html)...is there a way to pass through extra configuraiton to the codemirror initialization? This might let people have more flexibility over what happens.

Miniland1333:
Allows for passing through CodeMirror options specified here: [https://codemirror.net/doc/manual.html#config](https://codemirror.net/doc/manual.html#config)

Allows for enabling options such as readOnly, lineNumbers, and lineWrapping. Personally, I need all of these options for my use case.

closes #97

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,6 +92,9 @@ A full config script with defaults:
   //      // only apply the prompt stripping to cells matching this selector (optional)
   //      selector: '.sage-input',
   //    },
+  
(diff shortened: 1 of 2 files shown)


minrk:
Thanks!

PR #101, opened 2019-01-14
Convert site to Sphinx build

choldgraf:
Now that #100 is merged, this PR replaces the contents of gh-pages with the output of the sphinx build. Here's an example of the built site on my branch:

http://predictablynoisy.com/thebelab/

I'm not sure what's the best way to update this branch as I think this PR will create merge conflicts. I usually just force-push with [ghp-import](https://github.com/davisp/ghp-import), so lemme know if a PR isn't the right way to do this.


minrk:
Nice! Did you mean to make a PR against the `gh-pages` branch?

minrk:
Can this work on readthedocs? I generally try to avoid committing generated sphinx builds on gh-pages, since it infinitely bloats `git clone` of the repo.

choldgraf:
yep it can definitely work on readthedocs, I just don't have admin rights there so I don't believe I can set it up :-)

minrk:
Added you to https://readthedocs.org/dashboard/thebelab

choldgraf:
closing this since we're using RTD!

PR #104, opened 2019-01-17
Updating the requirements for alabaster_jupyterhub

choldgraf:
This updates the requirement for alabaster_jupyterhub to use the right name, it also adds a simple example to the index page so that users can see what thebelab is all about earlier.


choldgraf:
(sorry for the bumpiness of figuring out RTD...for this one I created a new conda environment and installed from doc-requirements and it worked, so I think this should work)

PR #105, opened 2019-01-17
Add ability to predefine preview output for cells

hugokerstens:
This PR makes it possible to render a div after the pre tag as a preview for the output of the cell.

When the option predefinedOutput has been set to true and the next element after the pre tag has data-output set to true, the html content of the next element will be rendered as a preview before execution.


src/thebelab.js
```diff
@@ -136,6 +137,7 @@ function renderCell(element, options) {
   // element should be a `<pre>` tag with some code in it
   let $cell = $("<div class='thebelab-cell'/>");
   let $element = $(element);
+  let $output = $element.next();
```
akhmerov: This isn't necessarily present, we should check if the element has a correct selector, probably `data-output` class by default.

minrk:
I think this makes sense! And is a continuation of #39. The only question I have is if we want to allow output to have any form other than an existing HTML element (i.e. if it should be allowed to be JSON jupyter output data). We can come to that later, though. Happy to merge after @akhmerov's comment is addressed.

src/thebelab.js
HugoDaBosss: If the next element is not present, then `$output` becomes an empty JQuery object, which results in `$output.data("output")` being `undefined`. This should not be a problem since no error is thrown and `$output.data("output")` still evaluates as `false` in the if statement.

akhmerov: Rather I meant when the next element is present, but it isn't `data-output`, but rather any other DOM entry.

HugoDaBosss: In that case `$output.data("output")` would evaluate to `undefined` and thus `false`. I agree with you that using a selector is better and addressed this in the commit 1e626a2.

akhmerov:
Very good! One last thing: I believe it would make sense to also make the output selector configurable like the input selector.

After that I think this is all set.

akhmerov:
Perfect, all is done from what I can tell. @minrk what do you think?

choldgraf:
One quick thought: I think this feature will be more discoverable if you link to it from the docs (https://github.com/minrk/thebelab/tree/master/docs)

hugokerstens:
The links to the examples in the docs (https://thebelab.readthedocs.io/en/latest/) do not seem to work (yet), but I added the link anyway for future purposes.

I switched the `demo-preview.html` to the CDN, although this does not work yet as it is not updated.

akhmerov:
@minrk, I just realized: `data-output` is likely an uncommon class to appear just on its own. Should this rather mean that `predefinedOutput` should default to true?

akhmerov:
@minrk @choldgraf I think this is good to go.

choldgraf:
Nice! Since @minrk suggested that we're good to merge once @akhmerov's comment was addressed, and since tests are happy and this comment was in fact addressed, I'm gonna merge this! Thanks @hugokerstens 🎉 🎉 🎉

akhmerov:
@choldgraf the docs seem to still load an old version of thebelab; time for a release?

choldgraf:
@akhmerov yep they do, unfortunately I'm not really sure how to cut a release :-/ not sure what @minrk and @nthiery 's plan for a release schedule is

PR #107, opened 2019-01-22
adding docs link in readme and adding screenshot

choldgraf:
Adds links to the RTD documentation to the README. Since the docs had a TODO item in there, I also added a simple screenshot.


choldgraf:
here's what the screenshot looks like:

minrk:
Cool!

PR #124, opened 2019-03-01
rely on @jupyterlab/codemirror to pull in codemirror dependency

minrk:
avoids duplicate codemirrors in the bundle when versions don't match exactly


choldgraf:
any reason not to merge this?

minrk:
Nope :)

PR #125, opened 2019-03-09
docs: load the latest thebelab version

akhmerov:


choldgraf:
Thanks so much! Nice catch :-)

PR #127, opened 2019-03-26
add working directory option

### LINKED ISSUE #114: Specify working directory in binder options
I have a github repository with subdirectories. In these subdirectories I have both thebe-enabled sites and some additional code, say `subdir/mycode.py`. When I use thebelab, binder always starts in the root directory of the repository, so that I cannot use the code in `subdir/mycode.py` which would be really convenient.

Is there a way to specify the (relative) working directory via the binderOptions? I have tried adding `?filepath=` to the `binderOptions.ref` string, which did not work.

I am not a web developer and don't know javascript, but from the console output:

```
Binder: Launching server...
thebelab.js:352:8
Binder phase: ready thebelab.js:340:8
Binder: server running at https://hub.mybinder.org/user/joergbrech-mode--und-simulation-fvgx3ovy/
thebelab.js:352:8
Starting WebSocket: wss://hub.mybinder.org/user/joergbrech-mode--und-simulation-fvgx3ovy/api/kernels/212d9d62-a028-4f36-ba61-905379ff7153 default.js:61
Kernel: connected (212d9d62-a028-4f36-ba61-905379ff7153) default.js:1432 
```
it seems like we could just request a binder server at `https://hub.mybinder.org/user/joergbrech-mode--und-simulation-fvgx3ovy/SUBDIR`, where `SUBDIR` could be specified via `binderOptions.workingDirectory` for instance...does that make any sense?

This issue is related to jupyter/jupyter-book#107.

joergbrech:
I don't know how I can test this, but I believe this would fix #114.

### DIFF
--- a/src/thebelab.js
+++ b/src/thebelab.js
@@ -67,6 +67,7 @@ const _defaultOptions = {
   binderOptions: {
     ref: "master",
     binderUrl: "https://mybinder.org",
+    workingDir: "",
   },
   kernelOptions: {},
 };
--- a/src/thebelab.js
+++ b/src/thebelab.js
@@ -426,7 +426,7 @@ export function requestBinder({
           es.close();
           resolve(
             ServerConnection.makeSettings({
-              baseUrl: msg.url,
+              baseUrl: msg.url  + "/tree/" + defaults.workingDir,
               wsUrl: "ws" + msg.url.slice(4),
               token: msg.token,
             })
--- a/README.md
+++ b/README.md
@@ -86,6 +86,7 @@ A full config script with defaults:
     binderUrl: "https://mybinder.org",
     // select repository source (optional). Supports Github(default), Gitlab, and Git
     repoProvider: "github",
+    workingDir: ""
   },
 
   // Options for requesting a kernel from the notebook server


choldgraf:
this seems reasonable to me - @minrk can you think of a reason that specifying a path to the binder build call will behave differently here?

src/thebelab.js
```diff
@@ -426,7 +427,7 @@ export function requestBinder({
           es.close();
           resolve(
             ServerConnection.makeSettings({
-              baseUrl: msg.url,
+              baseUrl: msg.url  + "/tree/" + defaults.workingDir,
```
minrk: The base URL here needs to be the base URL of the REST API. Where the kernel starts is not connected to this value.

minrk:
This definitely makes sense to do, but this PR doesn't quite do it. We'll need to switch to the [sessions api](https://jupyterlab.github.io/jupyterlab/services/modules/session.html#startnew), which adds a [path parameter](https://jupyterlab.github.io/jupyterlab/services/interfaces/session.ioptions.html#path) that `kernels.startNew` doesn't have. Passing `binderOptions.workingDir` to the path in `startNew` should get what we want here.

joergbrech:
I am closing this in favor of #131. @minrk Thanks for the feedback and @hugokerstens thanks for the new PR!

PR #131, opened 2019-04-25
Add ability to change path for binder

### LINKED ISSUE #114: Specify working directory in binder options

hugokerstens:
Change from `Kernel` to `Session` to support the `path` argument.
This adds the ability to specify `path` in `kernelOptions` so the current working directory is changed.

Fixes #114.

### DIFF
--- a/src/thebelab.js
+++ b/src/thebelab.js
@@ -7,7 +7,7 @@ if (typeof window !== "undefined") {
 }
 
 import { Widget } from "@phosphor/widgets";
-import { Session } from "@jupyterlab/services"
+import { Session } from "@jupyterlab/services";
 import { ServerConnection } from "@jupyterlab/services";
 import { MathJaxTypesetter } from "@jupyterlab/mathjax2";
 import { OutputArea, OutputAreaModel } from "@jupyterlab/outputarea";
--- a/src/thebelab.js
+++ b/src/thebelab.js
@@ -7,7 +7,7 @@ if (typeof window !== "undefined") {
 }
 
 import { Widget } from "@phosphor/widgets";
-import { Kernel } from "@jupyterlab/services";
+import { Session } from "@jupyterlab/services"
 import { ServerConnection } from "@jupyterlab/services";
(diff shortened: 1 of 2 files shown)


src/thebelab.js
```diff
@@ -7,7 +7,7 @@ if (typeof window !== "undefined") {
 }
 
 import { Widget } from "@phosphor/widgets";
-import { Kernel } from "@jupyterlab/services";
+import { Session } from "@jupyterlab/services"
```
slel: Missing trailing semicolon.

minrk: `npm run fmt` will fix any formatting

minrk:
Excellent, thanks!

PR #132, opened 2019-04-25
Remove / from .git suffix for binderUrl

hugokerstens:
This PR removes the / from the .git suffix for the binderUrl. When including the `/` the repository is not recognized, for example by a privately hosted GitLab server.

Now the following config DOES work, while without this PR it does not:
```
binderOptions: {
      repoProvider: "git",
      repo: "https://gitlab.example.com/user/repo",
      ref: "e6c970126f4e819e4a3eb717a86ba9fb5237fde3",
    }
```

### DIFF
--- a/src/thebelab.js
+++ b/src/thebelab.js
@@ -350,11 +350,6 @@ export function requestBinder({
     repo = repo.replace(/(^\/)|(\/?$)/g, "");
     // trailing / on binderUrl
     binderUrl = binderUrl.replace(/(\/?$)/g, "");
-    //add /.git if not present
-    if (!repo.endsWith(".git")) {
-      repo += ".git";
-      console.log("Added .git to repo name");
-    }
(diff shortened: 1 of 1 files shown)


src/thebelab.js
```diff
@@ -351,9 +351,9 @@ export function requestBinder({
     // trailing / on binderUrl
     binderUrl = binderUrl.replace(/(\/?$)/g, "");
     //add /.git if not present
-    if (!repo.endsWith("/.git")) {
```
minrk: I'm not sure when this got added, but I don't think we should be adding to the URL. Let's leave it exactly as the user specified.

choldgraf: as in, you think we should remove this whole `if` statement?

minrk: exactly

minrk:
Thanks!

PR #137, opened 2019-05-23
Drop cssnext bundle

SylvainCorlay:
`widgets.built.css` is the CSS of `@jupyter-widgets/controls` bundled with cssnext, which includes the polyfill for jupyterlab CSS variables (which we use for the classic notebook) and substitutes them by their values in all of the widgets CSS.

This PR makes thebelab use directly the bare CSS making use of the CSS variables that are already on the page in thebelab.

- Note that the tool that was used for this substitution was cssnext which is now deprecated.
- We still produce the old `widgets.built.css` in `@jupyter-widgets/controls` for backward compatibility.


SylvainCorlay:
For some reason, I don't manage to test this locally.

PR #160, opened 2019-08-06
Update everything

minrk:
this updates all dependencies for jupyterlab 1.0

- switches bundle analyzer to maintained `webpack-bundle-analyzer`
- updates shims and imports for new locations, unused additions (webpack tree-shaking really doesn't seem to work. Maybe this needs to be brought up upstream in jupyterlab, if there's something they can do to help).

The main tricky bit was the fact that webpack cannot transpile ES6 classes (i.e. jupyterlab dictates that all packages that *use* jupyterlab also target ES6, at least with webpack). This is a [babel bug, as far as I can tell](https://github.com/babel/babel/pull/8656). The fix was to bump minimum browser versions such that ES6 classes are okay. This notably means that IE and opera mini won't work (I'm not sure they ever did, but the definitely won't now).

closes #159 
closes #154
closes #153
closes #152
closes #151
closes #150
closes #149
closes #148
closes #146
closes #145 
closes #141


minrk:
@SylvainCorlay I was able to run the widgets example, so I think this works okay. But if you know of API changes that I may have missed, let me know.

SylvainCorlay:
@minrk this looks good to me.

I have ongoing work that goes a bit further, to make the `Output` widget and `ipywidgets.interact` work properly in thebelab, since it already does in voilà. My goal is to have the code for the two widget managers to converge. I think we can merge this and I will open a PR updating the widget manager when I am back from vacation.

minrk:
@SylvainCorlay sounds great, thanks!

PR #172, opened 2019-10-29
python mode without number

### LINKED ISSUE #96: Codemirror default language should be "python" not "python3"
I noticed that my code cells weren't being highlighted when I didn't specify `data-language`. I *think* that is because currently thebelab defaults to "python3"

https://github.com/minrk/thebelab/blob/c70aa40356e2918b990786da10f08fd889a20e29/src/thebelab.js#L225

but, I think instead this should just be "python".

### LINKED ISSUE #169: No python3 mode in CodeMirror
https://github.com/minrk/thebelab/blob/c58c2ee3ebe603c9a8413f815dcab06a657ef3e6/src/thebelab.js#L249

I guess there is no 'python3' mode in CodeMirror. Only the 'python' mode (default version is 3).

https://github.com/codemirror/CodeMirror/tree/master/mode/python/python.js

sergeyepimakhov:
Skip number to set the mode of the language in Codemirror. Closes #169, Closes #96

### DIFF
--- a/src/thebelab.js
+++ b/src/thebelab.js
@@ -246,7 +246,7 @@ function renderCell(element, options) {
   $cell.append(theDiv);
   Widget.attach(outputArea, theDiv);
 
-  const mode = $element.data("language") || "python3";
+  const mode = $element.data("language") || "python";
   const required = {
     value: source,
     mode: mode,


minrk:
Thanks!

PR #174, opened 2019-10-29
adding codemirror config page

choldgraf:
This adds a page to show how to configure codemirror for a page. It is a simple demo that changes the theme for codeMirror.

However, right now it doesn't seem to work! I'm not sure why, so we should figure that out...


minrk:
Great!

sharypovandrey:
I also can't figure out why CM styling doesnt work. Does somebody have the same issue?
Tried to import codemirror's css and js files, but didnt help.

PR #176, opened 2019-11-01
add css with standard jupyter ansi colours to make output colourful

sergeyepimakhov:
There was now issues about this, thought it can improve the output readability by making the widget colourful exactly in the same way like it was used in Jupyter.


minrk:
I imagine this is defined in the jupyterlab package, right? Perhaps it is a missing import? I'd rather import this than reimplement it, if we can.

sergeyepimakhov:
> I imagine this is defined in the jupyterlab package, right? Perhaps it is a missing import? I'd rather import this than reimplement it, if we can.

You were right. There was a stylesheet with ansi colours in jupyterlab/rendermime.

minrk:
Thanks!

PR #182, opened 2019-12-27
Feature/buttons in upper right corner

sergeyepimakhov:
Hi there, thought it's a good idea to place control buttons to the upper right corner of the cell, that could improve the usability. I combined also the both buttons Run and Restart into a button group. I did't find any issue concerning that. May be you can help me if you know any.

Cheers!

### DIFF
--- a/examples/index.html
+++ b/examples/index.html
@@ -16,9 +16,9 @@
       },
     }
   </script>
-  <!-- <script type="text/javascript" src="https://unpkg.com/thebelab@latest"></script> -->
+  <script type="text/javascript" src="https://unpkg.com/thebelab@latest"></script>
   <!-- or to use a local build: -->
-  <script type="text/javascript" src="../lib/index.js"></script>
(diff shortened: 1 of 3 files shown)


choldgraf:
Hey there, this looks interesting! could you post screenshots of what this changes visually? Or perhaps links to a website with these changes in it?

sergeyepimakhov:
image

choldgraf:
I think it looks nice, though this one probably merits a discussion to get agreement on the best location. For example, why is it better to have the buttons in the code blocks rather than below in the output block?

Another option would be to make the location of the buttons configurable, so users could choose what they wish. In that case it’d probably be best to make the current behavior the default though

sergeyepimakhov:
I find the idea to make the location of the buttons configurable very promising. Let me think about that.

choldgraf:
Sounds good - though it's probably best to get some consensus from the community before doing a ton of work in implementation. Not sure who else should weigh in, but probably @minrk at least

PR #192, opened 2020-04-24
Update deps

joelachance:
There were some vulnerabilities after running `npm audit`, so this fixes some of those. In addition, there was some unused code removed as well.

### DIFF
--- a/development.html
+++ b/development.html
@@ -59,7 +59,6 @@
         }
         .thebelab-cell {
             width: 50vw;
-            height: 20vh;
         }
     </style>
 </html>
--- a/package-lock.json
+++ b/package-lock.json
@@ -865,22 +865,419 @@
       }
     },
     "@jupyter-widgets/html-manager": {
-      "version": "0.18.3",
-      "resolved": "https://registry.npmjs.org/@jupyter-widgets/html-manager/-/html-manager-0.18.3.tgz",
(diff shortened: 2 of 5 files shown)


joelachance:
Ah, it looks like efforts were duplicated in PR #186. Either PR should be merged due to vulnerabilities.
@minrk @choldgraf

choldgraf:
Is there some way we can confirm that this still works as-expected? Did you test on your fork? Wanna rebase on master so that I can demo this with your nifty new developer preview code?

choldgraf:
I just tested this locally, and am noticing that some CSS rules don't seem to be applied properly. Here's the example from the main docs:

image

Do you think that's just because the other CSS rules from the documentation pages aren't loaded?

joelachance:
> Is there some way we can confirm that this still works as-expected? Did you test on your fork? Wanna rebase on master so that I can demo this with your nifty new developer preview code?

It seemed to work as expected, but I will rebase on master! On that note, I'm indifferent about removing the unused code here. It may be more of a risk than is necessary.

I do think its worth merging the dependency change due to vulnerabilities.

> I just tested this locally, and am noticing that some CSS rules don't seem to be applied properly. Here's the example from the main docs:

I'm going to look at this today-- I have a feeling this is due to some styling I put into the development.html file, and not Thebelab.

It'll be best for you to test it again quick after we make changes as I'm not 100% how everything should look & function just yet.

joelachance:
This got messy. I'm going to PR a change for the height fix, and then we might as well use this PR for the dependency update: https://github.com/minrk/thebelab/pull/186/files

I vote for not including the unused code removal commits for now.

PR #193, opened 2020-04-24
Feature/175 dev

joelachance:
Adds a development page in ./thebelab and contribution instructions in README.

I tried `webpack-dev-server`, however, I think the @jupyter package is too large as it doesn't rebuild after changes are made. This seems to work just fine, you simply have to reload the page after making changes.

Issue #183, opened 2019-12-29
Preview the documentation with CircleCI


choldgraf:
In #182 it would be useful to be able to preview the documentation change as a result of the feature change. This could be implemented with a quick CircleCI docs preview. I don't believe I have permissions to make the CircleCI connection, so @minrk (I believe you're the only one who can do this) could you either give me the admin permissions (in this repo) to do so, or set up a CircleCI build for this repo so we can configure it to preview the docs?

choldgraf:
@minrk can you add me as a CircleCI admin so that I can set this up? Otherwise I am not sure how to test things like #192 #193 #194

choldgraf:
closing this because we're now previewing docs with `readthedocs`!

PR #194, opened 2020-04-24
Loads CodeMirror Themes

### LINKED ISSUE #190: codeMirror themes not working
Following the documentation, we can arrive at this minimal example:

```
<!DOCTYPE html>
<html>
    <head></head>
    <body>
        <!-- Configure and load Thebe !-->
            {
                requestKernel: true,
                binderOptions: {
                repo: "binder-examples/requirements",
                codeMirrorConfig: {
                    "theme": "abcdef"
                },
            },
        }
        </script>
        <script src="https://unpkg.com/thebelab@latest/lib/index.js"></script>

        <button id="activateButton"  style="width: 150px; height: 75px; font-size: 1.5em;">Activate</button>
        <script>
            var bootstrapThebe = function() {
                thebelab.bootstrap();
            }

            document.querySelector("#activateButton").addEventListener('click', bootstrapThebe)
        </script>

        <pre data-executable="true" data-language="python">print("Hello!")</pre>
    </body>
</html>
```

It would be expected that the `abcdef` theme will work here, but it does not. I'm also noticing that it does not work in the docs example either. This issue was previously raised (#126), however, using the @latest tag in the CDN url doesn't seem to fix this.

joelachance:
Resolves #190 

Fixes our theme lookup. It also loads theme style as that isn't initially loaded. It only loads the theme that is given in the configuration.

### DIFF
--- a/docs/howto/cm_customize.rst
+++ b/docs/howto/cm_customize.rst
@@ -17,7 +17,7 @@ You can edit all of these attributes in a cell with the following thebelab confi
    </script>
 
 You can use any of `the available CodeMirror configurations <https://codemirror.net/doc/manual.html#config>`_.
-For example, the following configuration changes the CodeMirror theme:
+For example, the following configuration changes the `CodeMirror theme <https://codemirror.net/theme/>`_:
 
 .. code:: html
 
--- a/src/thebelab.js
+++ b/src/thebelab.js
@@ -255,8 +255,22 @@ function renderCell(element, options) {
       "Shift-Enter": execute,
     },
   };
+
+  // Gets CodeMirror config if it exists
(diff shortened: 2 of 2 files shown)


choldgraf:
I just merged #193 - wanna rebase so that I can test this out locally, since we don't yet have documentation preview builds working (#183)?  I am not a javascript guru, so it's hard for me to know if bugs are fixed without trying it out myself :-)

choldgraf:
ok I got antsy and pushed a rebase to your branch, just to try this out (I hope that was OK). It worked great! Thanks so much for the patch :-)

joelachance:
Of course! Glad this is in.

PR #195, opened 2020-04-28
Removes height styling

joelachance:


choldgraf:
looks great, thanks 👍

Issue #196, opened 2020-04-28
Compile thebelab so it can be loaded locally?


choldgraf:
I'm trying to think about how we can preview changes to Thebelab in pull-requests. Right now our documentation loads Thebelab from a CDN, but at least on *latest* we should be using "master" thebelab.

Is there a way for us to compile a dev build of thebelab that we can use in our documentation and in CI/CD testing to preview changes?

Maybe @joelachance has ideas for how this could be done?

joelachance:
@choldgraf How are deployments currently done? I'm assuming there is a number of steps to run the webpack build and upload the built file to the CDN, correct?

If I understand correctly, we'd like a preview environment for whatever is on master. There might be two ways to go here to create a staging environment:

1. Make a second CDN deployment with the latest/canary build, which could then be used in a second doc deployment. I'm assuming this might be a manual process, which we probably don't want. If this isn't manual, then this might be a good option.

2. Use a tool like Netlify to deploy a set of docs. This is a third-party integration we can connect to the repository. It will watch changes pushed to master, and can then run a given set commands. For the small number of people looking at latest builds (I'm assuming this is a small number, yes?), this should be free. You need to use quite a bit of bandwidth before they start asking for money, which I doubt we'd get remotely close to. This could run the webpack build and deploy the docs or the development.html (probably the docs).

I think the second option might be better unless we have some automated process for the first. Netlify will generate a non-obvious domain name (we can give it a custom domain, or <pick-our-own-subdomain>.netlify.com as well) for us to use as well. This would be the same every build.

Lastly, in order to flag the webpack build for the docs, I think we'd probably need to make a new entry in the Makefile and pass in the thebelab link, whether that's the webpack build for Netlify or the CDN address for an actual deployment.

Thoughts? Hopefully this all makes sense...

choldgraf:
For Netlify that should be straightforward, we could also use CircleCI to preview the documentation (that's what I've used in other projects). I am just not sure how to build a single local JS library that we can link to when building the documentation (so we aren't linking from the CDN). If you know how to do this then it should be easy to set up a "documentation preview" on any pull request

joelachance:
I'm wondering if I went too far earlier-- I'm going to PR a script shortly that will replace the script tag in the docs with one pointing at the local build (still need to fidget with this part). We would then put another command in the Makefile to run the script and build the docs, so we can preview it locally (not done yet).

[...]

PR #208, opened 2020-05-01
[WIP] Adds update_env script for staging deployment

joelachance:


joelachance:
This _might_ resolve #196, given an addition to the Makefile.

This is also another solution I came across that could work:
https://stackoverflow.com/questions/7250659/how-to-use-python-to-programmatically-generate-part-of-sphinx-documentation

Keep in mind we don't need to do this step if our 'staging' is using a simple HTML file like `development.html`.

I also had trouble building the docs (my environment), so I wasn't able to test this and get the right script src yet.

joelachance:
@choldgraf Wanted to ping you and see what your thoughts are here? No rush.

choldgraf:
Nice! Barring some better way of doing this (that I don't know about), this seems fine to me. We should just make sure to document it well so people know about it

joelachance:
Busy over here! 
I'll update this PR soon.

update-url.js
```diff
@@ -0,0 +1,22 @@
+const fs = require('fs');
+const glob = require("glob");
+
+const argIndex = process.argv.length-1;
+const env = process.argv[argIndex];
+
+const prod = '<script src="https://unpkg.com/thebelab@latest/lib/index.js"></script>'
+const staging = '<script src="LOCAL/index.js"></script>' // Need to change this
```
joelachance, line 8: We'll need to update this!

joelachance:
The last hiccup I'm having here is that I can't build the docs. I get an error complaining about `recommonmark`, but I can't seem to resolve that dependency on either of my systems. If we can update the local URL and make sure the build:docs npm script command works, I think that's the last blocker here!

choldgraf:
OK good to know. Hmmm, I can try to build it locally and see if it works then

joelachance:
@choldgraf, just a reminder about my comment in `update-url`, since it won't work 'out of the box': https://github.com/minrk/thebelab/pull/208/files#diff-15bcc910a34ad71aa4260aabea4a2673

update-url.js
minrk: Maybe the docs should always include a local build instead of sometimes referencing the CDN? Seems like each build of the docs should reference its own version of thebelab, since updates to the js could be incompatible with the examples in old versions of the docs.

joelachance: I might be misunderstanding, but are you suggesting we should default to the local build of Thebelab, or we have two separate builds?

My thought was that when there is a pull request, any changes to Thebelab would hopefully be reflected in the docs as well, so when Thebelab is deployed, new docs are also deployed in tandem.

I'm not sure how it works now, but that's how I assumed it might work :)

choldgraf: I think that @minrk is suggesting that the docs should always be building from *master* of thebelab js, and not use the CDN for this.

choldgraf: but, I can't tell whether @minrk is suggesting this PR not go through, or if it's a good step forward 🤔

PR #218, opened 2020-06-21
Updating documentation

choldgraf:
This makes a few updates to the documentation:

* Adds some configuration content that was brought up in recent issues
* Reorganizes some of the docs so that we have a single configuration page
* Adds getting started docs
* Adds circleci jobs to preview the docs artifacts
* Updates to the sphinx-book-theme

Here's how they look: https://28-84945484-gh.circle-artifacts.com/0/html/index.html

What do you think @minrk ?


minrk:
Awesome, thanks!

Issue #191, opened 2020-04-01
"request-kernel" event not being called


joelachance:
I'm attempting to run custom code in Thebelab, given the example in the docs:

```
thebelab.events.on("request-kernel")((kernel) => {
    // Find any cells with an initialization tag and ask ThebeLab to run them when ready
    kernel.requestExecute({code: "import numpy"})
}
```

I'm finding a couple of things here:

1. There are syntax errors here. We need a closing bracket at the end of the code block in both examples
2. `thebelab.events.on()` does not return a function

I assumed that the `on` function wanted a callback, but that also did not work.

Events code:
```
// events

export const events = $({}); // This is undefined when run in the browser, given we have jQuery as `$`
export const on = function() {
  events.on.apply(events, arguments); // since the above is undefined, this doesn't work
};
export const one = function() {
  events.one.apply(events, arguments);
};
export const off = function() {
  events.off.apply(events, arguments);
};
```

I don't fully understand what is happening above since I'm expecting events to have an `on` function assigned to it somewhere.  I assume the `on`, `one`, and `off` functions might be getting added to the `events` object in another file?

I'd be happy to submit a PR to fix some of this, but I'm not sure what the original intention was for the events feature. If there's any insight into how this should work, I propose the following:

- [ ] Update the events code
- [ ] Add some unit tests
- [ ] Update documentation

choldgraf:
As a general rule, I think that folks are friendly to PRs, though I believe the core maintainers (particularly @minrk ) are pretty unavailable right now due to lots of other work-related fires that have popped up. Just a note to set expectations around timing etc :-)   I would review your PR, but I don't think I have the javascript chops to make substantive contributions there

choldgraf:
also in case it's helpful, I have code that does something similar to what you're trying here: https://github.com/jupyter/jupyter-book/blob/master/jupyter_book/book_template/_includes/js/thebelab.html#L65 perhaps that can be used as inspiration?

joelachance:
@choldgraf , understood! I'm going to try and put some effort into this in the coming weeks, and hopefully all we need are some light code changes so a PR isn't so scary to merge.

Thanks for the response!

choldgraf:
That would be great - I'd love to see some love put into Thebelab ✨

PR #222, opened 2020-06-24
adding events docs

choldgraf:
This adds a page to discuss events in Thebelab, and how to hook into them.

Also removes CircleCI building for docs so that we can use readthedocs for PR previews

### DIFF
--- a/docs/events.rst
+++ b/docs/events.rst
@@ -1,0 +1,27 @@
+Event hooks in Thebelab
+=======================
+
+When Thebelab is launched (with ``thebelab.bootstrap``), it will emit a series
+of events corresponding to the state of the launch process. You can plug into
+these events to control the behavior on your page.
+
+To do so, use the ``status`` event within Thebelab, like so:
+
+.. code-block:: javascript
+
+       thebelab.on("status", function (evt, data) {
+        console.log("Status changed:", data.status, data.message);
+    });
+
+In the above code, the ``data`` object contains a collection of information about
+Thebelab, and ``data.status`` will reflect the current state of Thebelab. This will
+cycle between these states:
+
+* ``building``
+* ``built``
+* ``launching``
+* ``ready``
+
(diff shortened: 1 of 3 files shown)


docs/events.rst
```diff
@@ -0,0 +1,27 @@
...
+of events corresponding to the state of the launch process. You can plug into
+these events to control the behavior on your page.
+
+To do so, use the ``status`` event within Thebelab, like so:
+
+.. code-block:: javascript
+
+       thebelab.on("status", function (evt, data) {
+        console.log("Status changed:", data.status, data.message);
+    });
```
joelachance, line 14: Out of curiousity, does this work? I wasn't having luck previously getting events working, and I think it's an issue with the jQuery events implementation (issue #191)

choldgraf: I think it worked for me. This is the output in sphinx-thebelab:

image


here's where I'm using it:

https://github.com/executablebooks/sphinx-thebelab/blob/master/sphinx_thebelab/_static/thebelab.js#L39

choldgraf:
ahh I see - so should I change the event that I've shown here if it doesn't work?

joelachance:
Possibly! I haven't had luck getting events working myself, but curious if other people have had luck. The events example in the docs isn't a common JS pattern, plus there are syntax errors in there. Your example looks good, but just not sure if it works.

choldgraf:
oops - sorry if there are syntax errors. Are they in this PR? If so any chance you could suggest improvements and I'll merge them in?

joelachance:
Oh, I think the PR looks fine! There's some issues with the examples here: https://thebelab.readthedocs.io/en/latest/howto/initialize_cells.html, but I think this is covered in #191. Sorry for the confusion!

choldgraf:
ah ok - gotcha! thanks for the feedback, I'll merge this and it'd be awesome to see improvements to the other docs too!

PR #248, opened 2020-10-16
[DOC] add jsdoc to sphinx

minrk:
builds js api docs with jsdoc, just like we have for Python packages with sphinx-autodoc

mostly a placeholder, but good to get started

We need to add jsdoc-style docstrings to our `export function` items in order for them to be available here.


minrk:
rad, [it works](https://thebe--248.org.readthedocs.build/en/248/api.html)!

choldgraf:
very cool!

Issue #228, opened 2020-07-17
improve test coverage


minrk:
as in...any test coverage

The only tests we currently have in this repo are that the bundle builds at all. Not a big guarantee of success! We should have real tests that exercise putting thebe on the page, requesting kernels, running code, etc.; I just haven't had a chance to set this up. @blink1073 mentioned that jupyterlab has some helper packages for testing.

It would be awesome to get some pointers from @jupyterlab folks on test scaffolding for a jupyterlab-based browser package. We can probably also look at the right subset of their tests for reference.

Basic things to test, especially with known issues we've had:

- loading config from the page
- transforming divs to cells
- connecting to a kernel (local or with Binder? Binder will probably be a lot slower)
- rendering output, especially latex
- widgets
- syntax highlighting for multiple languages
- permutations of preloaded output, etc.

blink1073:
We have a `testutils` [package](https://github.com/jupyterlab/jupyterlab/tree/master/testutils) that has among other things:

- A JS wrapper for a Jupyter [Server](https://github.com/jupyterlab/jupyterlab/blob/b0a1ca62628a0b450ecec483c2f7aca6457d3b32/testutils/src/start_jupyter_server.ts#L34)
- Mock objects for [Kernels](https://github.com/jupyterlab/jupyterlab/blob/b0a1ca62628a0b450ecec483c2f7aca6457d3b32/testutils/src/mock.ts#L153) etc.
- Utilities for working with [notebooks](https://github.com/jupyterlab/jupyterlab/blob/master/testutils/src/notebook-utils.ts)

It also provides some standard `jest` config, that we use in our packages.

- [`coreutils`](https://github.com/jupyterlab/jupyterlab/tree/master/packages/coreutils) shows a fairly minimal example of the config and testing setup
- [`notebook`](https://github.com/jupyterlab/jupyterlab/tree/master/packages/notebook) is a kitchen sink: it uses all of the above mentioned utilities

For end-to-end testing, we currently use [`puppeteer`](https://github.com/jupyterlab/jupyterlab/blob/master/jupyterlab/chrome-test.js) to launch the application, wait for it to load, and make sure that no console errors or uncaught JavaScript errors are thrown.  The app is run from a [python](https://github.com/jupyterlab/jupyterlab/blob/master/jupyterlab/browser_check.py) entry point.

We have better e2e test support coming via [Galata](https://github.com/jupyterlab/team-compass/issues/77)

[...]

PR #257, opened 2020-10-17
Small changes to test setup/layout

stevejpurves:
Some initial changes to the test setup to help make expanding the test coverage easier and some docs.
Towards issue #228 

Its still not clear how to start the Jupyter Server but have started moving code so that we can test the bootstrap process next.

This could be considered draft until later today, where I will try add some more meaningful tests or if it's blocking someone, this can be brought in. Tests are green and changes I think are positive.


minrk:
Great! I'm happy to land this as-is to keep things moving.

Issue #255, opened 2020-10-17
Handle binders and kernels dying


akhmerov:
As far as I understand, thebe right now doesn't detect when kernels or binders stop being available because of culling for instance. Catching the proper errors and requesting kernels or binders again would be a way to improve on that.

akhmerov:
Indeed, since binder kills kernels really fast, the users are likely to face thebe UI breaking with this in the console:

![image](https://user-images.githubusercontent.com/2069677/96335240-9c0a3780-1077-11eb-91c4-997416615bc5.png)

Miniland1333:
When the kernel or binder dies, would it be acceptable to force the user to refresh their page through a notification before requesting a new one? An issue if it is restarted transparently/automatically is that state is lost and this ambiguity may be confusing to the users. Alternatively, a reconnect can flush all Thebe output but this would have to be done in a way that doesn't startle the user.

Hypothetical example:
User has 10 code blocks that need to be ran sequentially. While after running block 5, they wait long enough to timeout. If the kernel or binder is refreshed automatically, running block 6 would likely fail as none of the prerequisite code in blocks 1-5 had been ran in the new kernel. However, the Thebe output currently does not change so unless the page is refreshed or we code something in there would be a disconnect between the visual and what is happening in the kernel.

akhmerov:
I think it's reasonable if thebe shows an error message when it detects that the kernel died, and then restarts automatically. This is also similar to the jupyter behavior.

Miniland1333:
When this happens, does Jupyter automatically clear the output or in the scenario I outlined above would it throw an error?

Since at LibreTexts a lot of our users are using Thebe as their first exposure to Jupyter (and may not even know what Jupyter is under the hood), I want to ensure this is as clear as possible for them. Basically, I would like it to be obvious that in the event the state is reset with a kernel/binder restart that they would likely need to start from the beginning again. This of course only matters for pages where the code cells do depend on each other.

akhmerov:
Hm, come to think of it, just showing an error message "kernel died, please restart" might be better and easier—then it's clear that all state is lost. That would still require restart to catch the kernel dying (right now it doesn't neither).

choldgraf:
I think this also relates to #271 which could be the place where this information exists if the kernel dies

psychemedia:
I've wondered before about whether there is a need for a (polite) keep-alive routine, eg in the context of [RISE presentations running on Binder](https://github.com/damianavila/RISE/issues/524).

[...]

PR #259, opened 2020-10-17
request new kernel when kernel died

akhmerov:
Only handles the kernel part of #255

### DIFF
--- a/src/thebelab.js
+++ b/src/thebelab.js
@@ -184,13 +184,13 @@ function renderCell(element, options) {
     $("<button class='thebelab-button thebelab-run-button'>")
       .text("run")
       .attr("title", "run this cell")
-      .click(execute)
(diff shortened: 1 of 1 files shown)


choldgraf:
one quick thought here - is there some way for users to *know* that the kernel has been restarted? Perhaps we include a short warning message about it or something? Otherwise I fear folks will be confused when the state suddenly changes

akhmerov:
So far this is largely useless and it's more about me figuring out where to put things.

akhmerov:
As far as UI goes, there's going to be a message in the cell when the kernel is restarting. The problem I'm facing so far though is that a lot of functionality is only meant to be run once via bootstrap.

akhmerov:
closing for now, I'll need to read up more on this.

Issue #263, opened 2020-10-17
Security note about XSS in the docs


rowanc1:
I think it is worth noting that Thebe injects scripts and javascript into the page, and that is potentially a security threat for folks if they are using this in authorized pages (e.g. making API calls from the page). This can be known as self-xss, and could be a problem.

For example, executing:
```
%%html
<script>document.body.innerHTML = "";</script>
```

Will completely delete everything on the page. :)

I think a note about this for developers that it should be used on static webpages only, where there is no harm that can come to the user. If folks are game, I can take a stab at the wording for review.

moorepants:
This sounds like a good addition. Go for it.

rowanc1:
On it! 👍

choldgraf:
Lol that example is hilarious

I kind of wonder if we should sanitize certain kinds of outputs

rowanc1:
I think there are some other ways to do this as the library evolves, sandboxed iframes for example. But it in *much* more complicated. :)

We are thinking about this a lot at iooxa, because we are in an authorized environment.

choldgraf:
Yeah - definitely something to think about. An iframe wouldn't be _that_ much extra work would it? I wonder if juniper has any safety hatches in this way

rowanc1:
It actually is a bit more work, because the iframe needs to be hosted on a different domain than the docs. This is why a lot of sites have `githubusercontent.com` etc. You then need to communicate with the iframe with a post-message in our case (or have a stateful server side component, which is way more complicated).

choldgraf:
Ok I take it back that is much more complicated lol. We could try some hackier sanitation like special-casing %%html and other common ways to inject arbitrary html. Since we have access to the contents of the cell before it is sent to the kernel and the same on the return trip

rowanc1:
I also think that this is the *power* of thebe? Similar to Jupyter which really doesn't stop you from doing this either.

It is (I think) completely safe if run on a static site like public RTDs. I just don't want someone dropping this into their app and then having some weird thing and realizing it was thebe!

There are some of the components of thebe, like kernel management etc, that might be nice to have access to in a different way. Need to think on that a bit more. :)

Miniland1333:
For this security issue, does it rely on a bad-acting author writing the XSS code, or are there other ways that this could be carried out? This affects to what extent this is an administrative problems versus something that must be handled at a technical/code level.

[...]

PR #264, opened 2020-10-17
📚 DOCS: Security considerations page

rowanc1:
I have tried to be positive first about the benefits, and why we are doing this. I also gave some links to XSS and Self-XSS.

Let me know about the language, happy to update.

See #263


choldgraf (approved):
The content looks great to me!  (Sorry for slow responses, am on baby duty today)

docs/security.rst
```diff
@@ -0,0 +1,19 @@
...
+(:abbr:`XSS (Cross-Site Scripting)`) attacks,
+with the most likely case being an Self-:abbr:`XSS (Cross-Site Scripting)` attack.
+This happens when someone executes code that they do not understand, or is malicious.
+
+For example, the `%%html` and the `%%javascript` cell-magics in Jupyter can directly insert
+script tags into your page. They can potentially modify the DOM, make API calls on the users behalf,
+or run untrusted code.
+
+We recommend that you run Thebe in a static environment (e.g. ReadTheDocs or similar)
+that has no access to user credentials such as cookies or API keys.
```
minrk, line 19: 👍

docs/start.rst
```diff
@@ -57,4 +57,8 @@ Calling the bootstrap function is generally accomplished by connecting it to the
 .. tip::
 
    If ``bootstrap: true`` is in the Thebelab configuration, this will be triggered
-   automatically upon page load.
\ No newline at end of file
+   automatically upon page load.
+
+
+.. include::
+   security.rst
```
minrk, line 64: maybe add it to the index toc instead of included in another page?

choldgraf: I was thinking the same

rowanc1: Can do, I wasn't sure that we wanted it in the main TOC, so opted for here.

Will add it there now.

rowanc1: image


Thoughts?

choldgraf: Maybe just call the page "security considerations" and create a sub header for this more  in case there are other security considerations to document

rowanc1:
Updated! Also changed to MyST under immense pressure from @choldgraf 😜

choldgraf:
wohoo! thanks very much @rowanc1 ✨

PR #266, opened 2020-10-17
use localStorage to persist binder pods across pages

### LINKED ISSUE #224: Persist a server connection across pages
Many times, a user switches pages within the same domain and wants to run some more code. Right now, Thebelab will request a **new** binder session each time this happens, which puts burden on mybinder.org.

It would be great if we stored the Binder session information to re-use the session if it exists. Then a user would get super fast sessions in subsequent pageloads.

@ines kindly showed us where she does this in Juniper, I think we should follow the same pattern!

https://github.com/ines/juniper/blob/master/src/index.js#L57-L64

rkevin-arch:
This should resolve #224 (we can make a new issue for the persisting kernel (not binder pod) part if needed).

When Binder spawns a pod, the jupyter pod URL and token are saved in the browser's localStorage. When Thebe is activated and there is already a pod URL+token in localStorage, it will try to reuse the existing pod. If it fails, then it falls back to creating a new pod.

### DIFF
--- a/src/thebelab.js
+++ b/src/thebelab.js
@@ -396,22 +396,16 @@ export function requestBinder({
 
   return new Promise(async (resolve, reject) => {
     // if binder already spawned our pod and we remember the creds, reuse it
-    let spawned_pod_url = window.localStorage.getItem(
-      "thebe_binder_pod_url_" + url
-    );
-    let spawned_pod_token = window.localStorage.getItem(
-      "thebe_binder_pod_token_" + url
-    );
+    let existing_server = JSON.parse(window.localStorage.getItem("thebe-binder-" + url));
     if (
-      ignoreSavedPod === false &&
-      spawned_pod_url !== null &&
-      spawned_pod_token !== null
+      existing_server !== null &&
+      new Date().getTime() - new Date(existing_server.started).getTime() < 86400
     ) {
       console.log("Saved binder pod info detected");
       let settings = ServerConnection.makeSettings({
-        baseUrl: spawned_pod_url,
-        wsUrl: "ws" + spawned_pod_url.slice(4),
-        token: spawned_pod_token,
+        baseUrl: existing_server.url,
+        wsUrl: "ws" + existing_server.url.slice(4),
+        token: existing_server.token,
       });
       try {
         await Kernel.listRunning(settings);
@@ -471,10 +465,13 @@ export function requestBinder({
           es.close();
           try {
             // save the current connection url+token to reuse later
-            window.localStorage.setItem("thebe_binder_pod_url_" + url, msg.url);
             window.localStorage.setItem(
-              "thebe_binder_pod_token_" + url,
-              msg.token
+              "thebe-binder-" + url,
+              JSON.stringify({
+                url: msg.url,
+                token: msg.token,
+                started: new Date(),
+              })
             );
           } catch (e) {
             // storage quota full, gently ignore nonfatal error
--- a/src/thebelab.js
+++ b/src/thebelab.js
@@ -419,6 +419,7 @@ export function requestBinder({
           "Saved binder pod info seems to be invalid, respawning pod",
           err
         );
+        window.localStorage.removeItem("thebe-binder-" + url);
       }
     }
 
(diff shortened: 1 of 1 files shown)


minrk:
Great! I don't *think* persisting the kernel is as valuable as the server (kernel starts are fast, server starts with Binder are quite slow), so we can wait until someone requests that specifically.

rkevin-arch:
I agree. I also think persisting kernels across pages can be potentially confusing, since code run on the first page might affect stuff in the second page. If someone thinks there's a valid use case, we can deal with it then.

src/thebelab.js
```diff
@@ -436,6 +469,21 @@ export function requestBinder({
           break;
         case "ready":
           es.close();
+          try {
+            // save the current connection url+token to reuse later
+            window.localStorage.setItem("thebe_binder_pod_url_" + url, msg.url);
```
minrk, line 474: maybe a single item with the two fields as JSON? e.g. info 


```suggestion
            window.localStorage.setItem("thebe-binder-" + url, JSON.stringify({url: msg.url, token: msg.token, started: new Date()}));
```

then `started` can also be used to ignore saved data older than e.g. one day

src/thebelab.js
```diff
@@ -393,11 +393,44 @@ export function requestBinder({
...
+  return new Promise(async (resolve, reject) => {
+    // if binder already spawned our pod and we remember the creds, reuse it
+    let spawned_pod_url = window.localStorage.getItem(
+      "thebe_binder_pod_url_" + url
+    );
+    let spawned_pod_token = window.localStorage.getItem(
+      "thebe_binder_pod_token_" + url
+    );
+    if (
+      ignoreSavedPod === false &&
```
minrk, line 406: ignoreSavedPod is undefined?

src/thebelab.js
```diff
@@ -393,11 +393,44 @@ export function requestBinder({
...
   console.log("Binder build URL", url);
-  events.trigger("status", {
-    status: "building",
-    message: "Requesting build from binder",
-  });
-  return new Promise((resolve, reject) => {
+
+  return new Promise(async (resolve, reject) => {
+    // if binder already spawned our pod and we remember the creds, reuse it
+    let spawned_pod_url = window.localStorage.getItem(
```
minrk, line 399: Let's try to avoid kubernetes names in here, and call it e.g. "existing_server_url"

minrk (commented):
Awesome! This seems great. A few comments about details, but this looks very nice.

src/thebelab.js
rkevin-arch: Whoops, remnant of some other code that no longer exists. Removed.

rkevin-arch:
I've made some changes according to your comments. I also made it so the saved info is not used after 86400 seconds (a day). Probably should change it to an hour or leave it configurable. Let me know what you think.

choldgraf:
Really nice - thanks for this! I agree about not worrying about persisting kernels, since jupyter has a "1 kernel per notebook" mental model, I think most users will also assume "1 kernel per webpage"

rowanc1:
This works well for me!

For anyone curious on seeing or clearing the local storage, you can do that in Chrome here:
image

src/thebelab.js
```diff
@@ -393,11 +393,40 @@ export function requestBinder({
...
+      try {
+        await Kernel.listRunning(settings);
+        console.log("Saved binder pod is valid, reusing connection");
+        resolve(settings);
+        return;
+      } catch (err) {
+        console.log(
+          "Saved binder pod info seems to be invalid, respawning pod",
+          err
+        );
```
rowanc1, line 421:
```suggestion
        console.log(
          "Saved binder pod info seems to be invalid, respawning pod",
          err
        );
        localStorage.removeItem("thebe-binder-" + url);
```

rowanc1 (approved):
Have looked over the PR, and tested locally - including trying to screw up URLs and keys. :)

moorepants:
Sounds like this is good from the three reviewers. Gonna merge. Thanks @rkevin-arch!

choldgraf:
This is very exciting!

minrk:
This is rad! I added a little configuration to it in #280

PR #268, opened 2020-10-17
[DEV] use latest jupyterlab APIs

minrk:
updates everything except for webpack 5, which doesn't seem to work yet. This uses the packages from the latest jupyterlab 3.0 release candidate.

The changes are relatively small:

- use `KernelManager().startNew` instead of `Session.startNew`
- `appendToken` behavior is now opt-in instead of the default. This is required for cross-origin use cases like ours
- no longer accepts partial serverSettings, only filled-out by makeSettings

This seems to work just fine, but I don't want to introduce any conflicts with what other sprinters are working on


webpack.config.js
```diff
@@ -32,21 +32,23 @@ module.exports = {
...
+    // shim(/@lumino\/coreutils\/lib\/random/),
+    // shim out some unused lumino
     shim(
-      /@phosphor\/widgets\/lib\/(box|commandpalette|contextmenu|dock|grid|menu|scroll|split|stacked|tab).*/
+      /@lumino\/widgets\/lib\/(box|commandpalette|contextmenu|dock|grid|menu|scroll|split|stacked|tab).*/
     ),
-    shim(/@phosphor\/collections\/lib\/(bplustree).*/),
-    shim(/@phosphor\/(dragdrop|commands).*/),
+    shim(/@lumino\/collections\/lib\/(bplustree).*/),
+    shim(/@lumino\/(dragdrop|commands).*/),
```
minrk, line 44: The shims will need revisiting, as the bundle got even bigger again!

minrk:
This is working, and should no longer conflict with sprint contributions

choldgraf:
@minrk will this be an issue if JupyterLab isn't installed on the particular binder instance that a person requests? E.g. I think that Binder ships with JupyterLab 2.X right now, and I dunno when it'll switch to 3.x

minrk:
It shouldn't. This only means the jupyterlab *javascript* APIs that are pulled in as part of thebe itself. It should not be sensitive to the server-side version.

PR #269, opened 2020-10-17
📚 DOCS: Added a plotly example.

moorepants:
This functions for me if I include require.js:

image


moorepants:
This is ready. It matches the format of the ipyleaflet example and fully works.

choldgraf:
It looks great to me, thanks for the example! I think these will help people figure out how to set things up properly with JS libraries

Issue #237, opened 2020-08-21
Restart & Run All


mikaem:
Hi

First of all, thanks a lot for making a great piece of software:-) I am currently using jupyter book with thebe for both my [teaching](https://mikaem.github.io/MEK4300/content/intro.html) and for documentation of my [software](https://mikaem.github.io/shenfun-demos/content/intro.html).

I just have a small feature request. Would it be possible to add a Restart & Run All button for thebe/jupyter book? I use this feature in most of my notebooks all the time, and I think it would make my documentation much better and easier to use as well. 

Thanks a lot:-)

choldgraf:
Definitely - I think that this could be an addition somewhere around here:

https://github.com/executablebooks/thebe/blob/master/src/thebelab.js#L189

we'd need to figure out the right JupyterLab API call to trigger this as well.

manics:
It looks like it's not a single command, instead it's a dialog that restarts the session before running all cells:
https://github.com/jupyterlab/jupyterlab/blob/09b739d44d5dee6ad44b0243871c12b772b1b2d8/packages/notebook-extension/src/index.ts#L2299-L2309

rowanc1:
I am starting to look into this in ~30 min!

choldgraf:
Two quick thoughts on buttons etc:

At some point we may want to rethink the UIUX of buttons for thebe cells so they are not overloaded with buttons. Eg either make buttons configurable or use a drop-down for some?

rowanc1:
Right now I am going with something like this:

![image](https://user-images.githubusercontent.com/913249/96351533-2901c080-1079-11eb-86a4-2ece68431c99.png)

Happy to make changes once I get a bit more oriented in what I am actually doing ...... 🧐

choldgraf:
I think for now it makes sense to just go with that, UI/UX is probably another enhancement for the future and we might wanna think about it a little bit. We want to find the right balance between "things many people will want", and "fully recreating the notebook UI" lol

rowanc1:
Yep, this is the most minimally invasive approach for now. And mostly just me getting oriented in the lib.

moorepants:
Both "restart" and "restart and run all" should probably be displayed only once on a page. Would it be possible to have those two buttons only on the first cell? Maybe even above the first cell? Or should it be like the "activate" button, in that you only display it if you want the user to have that functionality.

rowanc1:
Can we tackle this outside of this small pr?

[...]

PR #270, opened 2020-10-17
✨ NEW: Restart and Run All buttons

rowanc1:
# Overview
This adds functionality to restart and run the kernel. It clears the outputs of each cell with a `"Waiting for kernel ... "` *before* the kernel is restarted (a slight difference from jupyter) but this is helpful as there is no other UI indicator that the button worked.

thebe-restart-and-runall

## Updates

* I have changed the return of `renderCell` to return `{cell, execute, setOutputText}` which allows you to use them in other contexts.
* There is now a `thebelab.cells` which is an array of the above.

See #237

### DIFF
--- a/src/thebelab.js
+++ b/src/thebelab.js
@@ -192,6 +192,12 @@ function renderCell(element, options) {
       .attr("title", "restart the kernel")
       .click(restart)
   );
+  $cell.append(
+    $("<button class='thebelab-button thebelab-restartall-button'>")
+      .text("restart & run all")
+      .attr("title", "restart the kernel and run all cells")
+      .click(restartAndRunAll)
+  );
   let kernelResolve, kernelReject;
   let kernelPromise = new Promise((resolve, reject) => {
     kernelResolve = resolve;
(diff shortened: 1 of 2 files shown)


choldgraf:
This looks good to me! Note that I added another cell to the `development.html` example in order to be able to demo this locally (which worked!), and then when I pushed to your branch github informed me that apparently another example has already been added to that file on `master`, so I just merged in that part of the master branch 👍

choldgraf:
we probably want some kind of indication to users that a kernel has been restarted. Maybe we clear the outputs or put a little "loading" SVG or something like that. (doesn't need to be in this PR)

rowanc1:
Ha yeah, I was seeing @minrk working on that.

Do you think this is good to merge in?

src/thebelab.js
```diff
@@ -237,10 +247,24 @@ function renderCell(element, options) {
   function restart() {
     let kernel = $cell.data("kernel");
     if (kernel) {
-      kernelPromise.then((kernel) => {
-        kernel.restart();
+      return kernelPromise.then(async (kernel) => {
+        await kernel.restart();
```
choldgraf, line 251:
```suggestion
        setOutputText("Restarting kernel...")
        await kernel.restart();
```

is adding an indication for a restarting kernel as simple as this?

rowanc1: That will only work for the current cell I believe.

choldgraf:
this looks good to me and testing it locally worked great, so I'll merge this in and we can continue iterating on the buttons / layout / etc in subsequent issues! Thanks @rowanc1 ✨

PR #272, opened 2020-10-17
📚 DOCS: header fix on contribute page

choldgraf:
There are multiple top-level headers on the "contribute" page. This makes them H2 so that we don't have multiple "titles" for the page.


rowanc1 (approved):
Looks good to me!

PR #273, opened 2020-10-17
📚 DOCS: Go through docs/strings & change ThebeLab -> Thebe

rowanc1:
This should be a non-invasive change to refresh some of the documentation when referring to Thebe.

I have updated some code, but only comments and strings that have no other effects.

### DIFF
--- a/examples/status_field.html
+++ b/examples/status_field.html
@@ -1,7 +1,7 @@
 <html>
 <head>
   <meta HTTP-EQUIV="Content-Type" CONTENT="text/html; charset=UTF-8">
-  <title>Thebe Lab example: status button, styling, ...</title>
+  <title>Thebe example: status button, styling, ...</title>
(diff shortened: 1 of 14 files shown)


examples/status_field.html
```diff
@@ -27,15 +27,16 @@
   <script src="thebe_status_field.js" type="text/javascript"></script>
   <link   href="thebe_status_field.css" rel="stylesheet" type="text/css"/>
 
-  <script><!-- use window.onload to defer the execution until the page is loaded !-->
+  <script>
+    // use window.onload to defer the execution until the page is loaded
```
rowanc1, line 31: This is the only real "code" change. Just updated it while I am here.

choldgraf:
nice! thanks very much for this :-) step by step we are closer to the name change being complete!

rowanc1:
Mindless task, but it makes me walk through all sorts of files to get a lay of the land. :)

PR #274, opened 2020-10-17
Added per-cell readOnly option

Miniland1333:
By adding `data-readonly="true"` to a cell, the contents of that cell cannot be as easily modified. Testing indicates that CodeMirror actually also prevents modifications to the HTML after instantiation to affects this as well, which significantly raises the barrier for tampering. For examples/demos, this would be useful if the user is not expected to modify the code.

Example usage:
`<pre data-executable="true" data-language="python" class="pre-element" data-readonly="true">print("Hello!")</pre>`

Note that this does not interfere with the existing ability to set all cells on a page to read-only using `{"codeMirrorOptions":{"readOnly":true}}`


src/thebelab.js
```diff
@@ -248,13 +248,17 @@ function renderCell(element, options) {
   Widget.attach(outputArea, theDiv);
 
   const mode = $element.data("language") || "python";
+  const isReadOnly = $element.data("readonly") || false;
```
Miniland1333, line 251: Should `data-readonly=""` and `data-readonly` (which are viewed equivalently in JS) be considered readOnly or not? This determines if this should be `||` falsy or `??` nullish coalescing.

minrk: The convention I can find suggests that for booleans like this, it should be `<pre data-readonly>...` to be considered true, so the only false values ought to be `undefined` and (maybe) explicit `false`:

```html
<pre> # false
<pre data-readonly='false'> # false
<pre data-readonly> # true
<pre data-readonly='true'> # true
```

I *would* say that we should follow the boolean-only convention and use

```python
const isReadOnly = $element.data("readonly") !== undefined;
```

but the fact that we already have `data-executable="true"` (which I now realize was maybe a mistake), suggests we should support true/false to be consistent with ourselves.

In the future, we should probably add support for this no-right-hand-side convention in data-executable as well, to be consistent.

docs/minimal_example.rst
```diff
@@ -144,5 +144,11 @@ Here's the HTML for the cell above:
...
    plt.plot(x, np.cos(x))
    </pre>
+   
+And here's an example where the contents cannot be modified once instantiated:
 
-For more examples, check out :ref:`more_examples`.
\ No newline at end of file
+.. raw:: html
+
+   <pre data-executable="true" data-language="python" data-readonly="true">print("My contents cannot be changed!")</pre>
```
minrk, line 152: Maybe to encourage adoption the more idiomatic convention:

```suggestion
   <pre data-executable="true" data-language="python" data-readonly>print("My contents cannot be changed!")</pre>
```

minrk (commented):
Terrific!

Miniland1333:
In this particular case, I think it is important that explicit false is accepted as it can be used to selectively override `{"codeMirrorOptions":{"readOnly":true}}`. That way, an author could use the codeMirror options to make them readonly by default and explicitly allow editing for certain blocks. I'll do some quick modifications/testing tomorrow to ensure the code can accommodate this.
My guess is that it will look like:
`const isReadOnly = $element.data("readonly") !== undefined && $element.data("readonly") !== "false";`

Miniland1333:
I've updated this code to handle the explicit false case. Do you think it would be a good idea to write tests for this based on the different cases?

moorepants:
@Miniland1333 Looks like you'll need to merge in master and resolve the conflicts.

Miniland1333:
> @Miniland1333 Looks like you'll need to merge in master and resolve the conflicts.

Okay, the conflict has been resolved.

moorepants:
> Do you think it would be a good idea to write tests for this based on the different cases?

Yes, this is a good thing. One of the goals is to have the software properly and thoroughly tested.

Miniland1333:
> > Do you think it would be a good idea to write tests for this based on the different cases?
> 
> Yes, this is a good thing. One of the goals is to have the software properly and thoroughly tested.

I'm currently writing and testing some tests for Jest so hold off on the review until I get this working. Related to conversation at #228

Miniland1333:
@minrk I have finished writing the tests in Jest and it is quite nice. These tests can be run using `npm run test:jest`. When you have time, can you as well as a second person review this PR due to the enlarged scope? I will also ask the JupyterTeam to take a look at this and see if they can break my tests.

Things to check:
===
JEST
---

- See how I am using HTML fixtures to set up testing scenarios.
- See how I have set up the test/readonly.test.js file with my tests.
- Run JEST tests. This has a really nice Webstorm integration for progress reporting.
image

Thebe
---

- $element's id is now transferred to the replacement $cell.
- read-only attribute is transferred to $cell and the inner codeMirror.

sandertyu:
I went ahead and ran `npm run test:jest` as requested and here was my output;

image

I'm not sure how useful these test files are, but they do pass quite fine.

narrator0:
@Miniland1333 I think the work is great and the tests do their job. The only thing that bothers me is that I have to look at both the fixture HTML file and the test file at the same time to understand what they are testing. It might be easier for people to review if we can have all the settings put in `readonly.test.js`.

We can have one general fixture HTML that only have the bare minimum. Then we can change the thebe config in the test cases and run `boostrap` manually. I think this will make code review easier and also allow us to not create one HTML file per test case.

moorepants:
Positives from @narrator0 and @stevejpurves in #228, so merging.

docs/minimal_example.rst
```diff
@@ -144,5 +144,11 @@ Here's the HTML for the cell above:
    plt.plot(x, np.sin(x))
    plt.plot(x, np.cos(x))
    </pre>
+   
+And here's an example where the contents cannot be modified once instantiated:
```
choldgraf, line 148: Hey all - sorry I didn't get to this earlier. I think it's fine to merge but we really need to document things better than this if we add new features.

Also @moorepants could you please check out [the executablebooks guidelines for merge commits](https://executablebooks.org/en/latest/contributing.html#commit-messages) and format the squash-and-merge message accordingly? Thanks!

moorepants:
I'm not a big fan of squashing. Is that a requirement?

stevejpurves:
having atomic commits along master, can be really helpful if you need to back out a change. 1 PR per issue and 1 commit per PR does make that easier to achieve (when possible), so I see the sense in squash-and-merge when things go into master. 

sounds like you see some negatives though @moorepants? have yo hit difficulties in the past?

moorepants:
Yes, it's an annual debate over at SymPy where I also contribute. I lean on the side that it adds unnecessary overhead when merging is just fine and requires less git kung fu when merge conflicts arise. Our standard, but debated, practice is to avoid squashes and rebasing unless necessary. I can do whatever the established practice of this community is, but I tend to like minimizing overhead as long as it doesn't compromise things deemed important.

choldgraf:
It's not a hard requirement if there are two or more distinctly atomic changes being made in a PR (in that case, I believe folks usually try to rebase those commits onto master). But in general we've found that 1 PR -> 1 atomic contribution, and so we use squash merges along with standardized commit messages (e.g. `✨ NEW: Added XYZ new feature`) to keep the commit history clean and readable

minrk:
I'm not personally a fan of squashing (or enforcing change classification in commit messages), but I'll defer to the org hosting this repo which appears to be stricter than I'm comfortable enforcing.

choldgraf:
I would love to hear thoughts about why you don't like doing this. Is it primarily because of extra maintenance burden? Or you're worried about lumping changes into a single commit that obscures what exactly happened?

minrk:
I don't have a strong objection, but it is mainly about maintenance overhead. In particular, the classification task if not the actual reformatting of the message. It's more work to think about and classify a PR, which can get harder if a PR is a larger project adding a feature, which in turn requires fixing a bug (or vice versa) and documenting the changes while you're at it. What's the right format there? If it's not important, what's the benefit of doing it? To be clear, I don't object to doing it at all, or even recommending it, I just don't love *requiring* it. I actually don't know if I'm merging PRs correctly on this repo according to executablebooks guidelines, which makes me uncomfortable. As something of a "guest maintainer" here, I don't know how strict I'm supposed to be, and should probably stop merging PRs here.

I do strongly object to requiring *contributors* to follow the strict [commit message](https://executablebooks.org/en/latest/contributing.html#commit-messages) rules, though they are AOK as a *guide*. That's I guess the main benefit of squash & merge, since it brings this burden onto maintainers away from contributors. I mainly think that lots of effort going into a 'clean git history' doesn't have benefits outweighing the costs and frustration related to enforcing a strict process, whether from a contributor who is "contributing wrong" or a maintainer having to do significant extra processing and work beyond accepting changes.

To me, preserving merge commits has all the benefits of squashing (all PRs can be seen by listing only merge commits), plus eliminating the overhead of deciding if/how to classify a PR and whether squashing is appropriate or not, plus preserving commit information in larger  PRs without every having to ask "what's the right way to merge this?" to destroy just the right amount of information. I do think it's better to use squash *or* merge and never rebase, since that does detach things from the PR and time, both. I also think the "interactive rebase" recommendation at the end of [this documentation](https://executablebooks.org/en/latest/contributing.html#merging-pull-requests) is not to be recommended except in extreme cases, e.g. lots of "fixup" and "debug" commits.

If the formatting is used for things like generating history/changelog stuff, I think using Issue tags is better, in large part because it can be fixed after-the-fact (tags and titles can be edited after merge, commit messages can't).

moorepants:
Nice put @minrk. I have the same opinion on all of those aspects.

choldgraf:
So, I want to highlight that the beginning of the "development conventions" page starts with

> This page outlines conventions and best practices for development and maintenance across all repositories in the EBP organisation, to help the community make the best tools possible. These are mostly guidelines, not rules. Use your best judgment, and feel free to propose changes to this document in a pull request.

AKA, those aren't strictly enforced rules, but just general guidelines that the EBP project follows. In this case I was recommending following the "squash and choose a commit message" approach because I've personally found it to be a lightweight and useful way to keep the commits "atomic" (and as you mention, this is something that the maintainer does just before hitting 'merge', not something that users generally do, though some adopt this practice on their own).

If you have suggestions for how to highlight that first paragraph in a way that's more obvious, then please let me know how we can improve this...

PR #279, opened 2020-10-19
show an error message when the kernel is dead

akhmerov:
This doesn't hook the kernel up again, and only shows an error message. But hey—still better than silently dumping exceptions into the console. :)


minrk:
Progress!

PR #280, opened 2020-10-19
✨ NEW: make savedSession behavior configurable

minrk:
adds binderOptions.savedSession configuration to give users a little control over #266

- toggle behavior off entirely with `savedSession.enabled`
- set max age in seconds with `savedSession.maxAge`
- specify storage prefix with `savedSession.storagePrefix`

and reorganize the code a little bit to make the flow a little easier to manage with all the different stages to abort, including JSON parse errors, etc.

### DIFF
--- a/README.md
+++ b/README.md
@@ -75,10 +75,19 @@ A full config script with defaults:
   // Options for requesting a notebook server from mybinder.org
   binderOptions: {
     repo: "minrk/ligo-binder",
+
+    // only repo is required, the rest below are defaults:
     ref: "master",
     binderUrl: "https://mybinder.org",
     // select repository source (optional). Supports Github(default), Gitlab, and Git
     repoProvider: "github",
+    savedSession {
+      // if enabled, thebe will store and try to re-use
+      // connections (with credentials!) to running servers
+      enabled: true,
+      maxAge: 86400, // the max age in seconds to consider re-using a session
+      storagePrefix: "thebe-binder-",
+    }
   },
 
   // Options for requesting a kernel from the notebook server
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ A full config script with defaults:
     binderUrl: "https://mybinder.org",
     // select repository source (optional). Supports Github(default), Gitlab, and Git
     repoProvider: "github",
-    savedSession {
+    savedSession: {
       // if enabled, thebe will store and try to re-use
       // connections (with credentials!) to running servers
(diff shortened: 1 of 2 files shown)


moorepants:
@rkevin-arch Could you review this?

README.md
```diff
@@ -75,10 +75,19 @@ A full config script with defaults:
   // Options for requesting a notebook server from mybinder.org
   binderOptions: {
     repo: "minrk/ligo-binder",
+
+    // only repo is required, the rest below are defaults:
     ref: "master",
     binderUrl: "https://mybinder.org",
     // select repository source (optional). Supports Github(default), Gitlab, and Git
     repoProvider: "github",
+    savedSession {
```
rkevin-arch, line 84: `savedSession: {`?

rkevin-arch:
The code looks great, gonna test it locally in a bit

src/thebelab.js
```diff
@@ -424,34 +433,69 @@ export function requestBinder({
...
+      return;
+    }
+    let storedInfoJSON = window.localStorage.getItem(storageKey);
+    if (storedInfoJSON == null) {
+      console.debug("No session saved in ", storageKey);
+      return;
+    }
+    console.debug("Saved binder session detected");
+    let existingServer = JSON.parse(storedInfoJSON);
+    let timestamp = new Date(existingServer.timestamp);
```
rkevin-arch, line 449: This should be `existingServer.lastUsed` rather than `existingServer.timestamp`

src/thebelab.js
```diff
@@ -424,34 +433,69 @@ export function requestBinder({
...
+    if (new Date().getTime() - timestamp.getTime() < savedSession.maxAge) {
```
rkevin-arch, line 450: 1. The if statement should trigger if the time difference is _bigger_ than the max age
2. This is my fault, but upon googling the `getTime()` function returns milliseconds rather than seconds, so we should probably divide by 1000

rkevin-arch:
Other than these, it looks pretty good to me

minrk:
@rkevin-arch thanks for your review! I think it's all set now.

choldgraf:
This looks great to me - thanks @rkevin-arch for the review. I'm +1 on this once you are.

choldgraf:
OK given that review comments are addressed and tests are happy, I'm just gonna merge this one and we can iterate in another PR if there are follow-ups needed!

PR #281, opened 2020-10-19
📚 DOCS: Moved minimal example into docs/examples & fixed url.

moorepants:
image


moorepants:
@choldgraf This is good to go. Just needs a review.

choldgraf:
nice! I like the new sub-section of `examples` 👍

PR #282, opened 2020-10-19
Render Cell Tests

stevejpurves:
DONE:
 - readOnly integration tests have been moved to e2e with their own jest config
 - unit style tests added and passing, with their own jest config that transpiles thebe and dependencies
 - updated ci to run both sets


minrk:
Awesome! Thanks for adding this. Looks like the only failure is the linter: can you run `npm run fmt` and commit the result?

stevejpurves:
@minrk linted. I had the pre-commit hook in place, no idea why it didn't run.

stevejpurves:
@minrk did you see my comment / question on https://github.com/executablebooks/thebe/issues/228??

stevejpurves:
@minrk i've moved this forward. We can either put this up for review/merge as is? 
putting the jupyter server work (which i'll do next) in a separate PR, or wait.

package.json
```diff
@@ -77,11 +78,16 @@
...
-    "puppeteer": "^5.4.0",
+    "puppeteer": "^5.4.1",
     "style-loader": "^2.0.0",
     "url-loader": "^4.1.1",
     "webpack": "^4.0",
     "webpack-bundle-analyzer": "^3.9.0",
     "webpack-cli": "^4.1.0"
+  },
+  "dependencies": {
+    "identity-obj-proxy": "^3.0.0",
```
minrk, line 89: I don't think there should be any dependencies added here. Are these only test dependencies?

package.json
```diff
@@ -11,9 +11,10 @@
     "fmt": "prettier --trailing-comma=es5 --write *.js src/*.js examples/*.js test/*.js",
     "serve": "http-server -c-1 -a 127.0.0.1 -o development.html",
     "develop": "concurrently \"npm run build:watch\" \"npm run serve\"",
-    "test": "karma start --single-run --browsers ChromeHeadless karma.conf.js",
-    "test:jest": "jest test.js",
-    "test:watch": "karma start --auto-watch --browsers ChromeHeadless karma.conf.js"
+    "test": "jest",
```
minrk, line 14: Does this mean karma and mocha can be removed as dependencies?

minrk (commented):
This is really nice! I only had a question about the dependencies, but the test changes look great.

Miniland1333:
@stevejpurves, since #297 has been merged in to master would you be able to update from master to resolve these conflicts? And yes, I think you/we should remove the Karma and Mocha dependencies to clean that up.

package.json
stevejpurves: yes! i need to clean up good point

stevejpurves:
Hi @minrk are we ok to merge this one?

moorepants:
It seems you've addressed @minrk's comments. I'll merge near the end of my work day if there is no more feedback.

moorepants:
> Does this mean karma and mocha can be removed as dependencies?

Maybe this needs to be done in a new PR?

PR #283, opened 2020-10-20
Docs: create cytoscape example

celine168:
Documents the use of ipycytoscape, based on @moorepants's PR for documenting ipythreejs.


moorepants:
Cool. Add this to the toctree in the index.rst file.

moorepants:
This works for me locally:
image

docs/ipycytoscape_example.rst
```diff
@@ -0,0 +1,137 @@
+====================
+IPyCytoscape Example
+====================
+
+Thebe can display output from ipycytoscape_, which could visualize graphs using Cytoscape.js_.
```
moorepants, line 5: Mention the license of the code you copied from ipycytoscape (if you did) like we did in the other examples.

docs/ipycytoscape_example.rst
```diff
@@ -0,0 +1,137 @@
...
+
+Setup
+=====
+
+Be sure to load require.js before any of your thebe activation code so that the
+Cytoscape visualizations can function:
+
+.. code-block:: html
+
+   <script src="https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.4/require.min.js"></script>
```
moorepants, line 18: Did you verify that this is required? (no pun intended)

moorepants:
@celine168 If you make the two updates, this can be merged.

docs/ipycytoscape_example.rst
celine168: Yes, this is required, or the widget fails to load.
image

This is from the same code except removing require.js.

docs/ipycytoscape_example.rst
celine168: Added the mention to the BSD-3 Clause license.

docs/ipycytoscape_example.rst
moorepants: Thanks. It would be nice to know which ones require require.js. Seems like all of the ones we've tested so far.

moorepants:
LGTM!

PR #286, opened 2020-10-23
📚 DOCS: add read-only docs

choldgraf:
follow-up to #274 

could @Miniland1333 or @moorepants confirm that these instructions are correct?

### DIFF
--- a/docs/configure.rst
+++ b/docs/configure.rst
@@ -134,3 +134,17 @@ The below code cell demonstrates this theme:
 
 The above code should be styled according to the
 `CodeMirror abcdef theme <https://codemirror.net/demo/theme.html#abcdef>`_.
+
+
+Mark a code cell as read-only
+=============================
+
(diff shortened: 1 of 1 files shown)


moorepants:
LGTM

choldgraf:
great, thanks!

Miniland1333:
The only thing is that do you think we should mention the more specific use case of readOnly as default? If `{codeMirrorOptions:{readOnly:true }` is set in the Thebe config it makes cells readonly by default. Then, `data-readonly="false"` can be used to make certain cells explicitly editable. This is useful for my use case, but I had problems trying to document this without making it seem overly complex. This is also why I had to have extensive tests for such a relatively simple feature.

choldgraf:
ah yes totally, I think it's a good idea to document how to enable this for all cells 👍 

want to make a PR giving it a shot and I'm happy to review?

Miniland1333:
Continued in PR #287

PR #287, opened 2020-10-24
📚 DOCS: Added language about read-only codeMirror default

Miniland1333:
- Mentions how read-only can be set as the default in codeMirrorConfig, as well as how to selectively override this using `data-readonly="false"`.
- Swapped some code blocks to `code:: javascript` for better highlighting.

Due to the higher complexity of this scenario, I need others to look this over and ensure it is straight forward and understandable.

Relates to #286 and #274

### DIFF
--- a/docs/configure.rst
+++ b/docs/configure.rst
@@ -82,7 +82,7 @@ CodeMirror is the tool used to convert your code cells into editable cells.
 It has a number of configuration options, such as theming and syntax highlighting.
 You can edit all of these attributes in a cell with the following thebe configuration:
 
-.. code:: html
+.. code:: javascript
 
 
    // Additional options to pass to CodeMirror instances
@@ -91,7 +91,7 @@ You can edit all of these attributes in a cell with the following thebe configur
 You can use any of `the available CodeMirror configurations <https://codemirror.net/doc/manual.html#config>`_.
 For example, the following configuration changes the `CodeMirror theme <https://codemirror.net/theme/>`_:
 
-.. code:: html
+.. code:: javascript
 
    codeMirrorConfig: {
        theme: "abcdef"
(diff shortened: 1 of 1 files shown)


docs/configure.rst
```diff
@@ -139,12 +139,27 @@ The above code should be styled according to the
...
 
 .. code-block:: html
 
    <pre data-executable data-readonly>print("I cannot be modified")</pre>
 
-Users will not be able to modify the code when Thebe is activated, though they can still
+Users will not be able to modify the code once Thebe is activated, though they can still
 press the "run" button to see the outputs.
+
+Alternatively, you can use the codeMirrorConfig in order to set the default as read-only for all cells:
```
choldgraf, line 152:
```suggestion
**To set all cells as read-only by default**, use the following `thebe` configuration:
```

docs/configure.rst
```diff
@@ -139,12 +139,27 @@ The above code should be styled according to the
...
+
+.. code:: javascript
+
+   codeMirrorConfig: {
+       readOnly: true
+   }
+
+And when read-only as the default, you can still override this setting for individual cells using `data-readonly`:
```
choldgraf, line 160:
```suggestion
This uses codeMirror to mark all cells as read-only. If you are using this setting and would like to
manually mark cells as editable, you can override the codeMirror configuration for individual cells using `data-readonly="false"`. For example:
```

choldgraf (commented):
This looks good! Thanks for the explanation. Two quick suggestions

docs/configure.rst
```diff
@@ -139,12 +139,28 @@ The above code should be styled according to the
...
+**To set all cells as read-only by default**, use the following `thebe` configuration:
+
+.. code:: javascript
+
+   codeMirrorConfig: {
+       readOnly: true
+   }
+
+This uses codeMirror to mark all cells as read-only. If you are using this setting and would like to
+manually mark individual cells as editable, you can override the codeMirror configuration for a cell using `data-readonly="false"`. For example:
```
Miniland1333, line 161: I can't figure out why \`data-readonly="false"\` isn't rendering as `data-readonly="false"` here for inline code. Any ideas? I think this is the last outstanding issue for this PR.

choldgraf, line 161:
```suggestion
manually mark individual cells as editable, you can override the codeMirror configuration for a cell using ``data-readonly="false"``. For example:
```

choldgraf: ah my bad - it was because I used a single backtick and rST requires two backticks by default

choldgraf:
This looks great, thanks very much for the follow-up @Miniland1333 ! If tests are happy then I'll 🚢it!

Issue #185, opened 2020-01-30
Navigation icons missing when integrating matplotlib widgets into Thebelab


tannavee:
I am trying to integrate the matplotlib widgets into Thebelab, but the icons are missing from the navigation bar (screenshot 1). I was wondering how I would go about resolving this. The nav bar should look like this (screenshot 2). I appreciate any input.

![image](https://user-images.githubusercontent.com/53124295/73410310-abadda00-42b6-11ea-805b-367dcaa652c9.png)

![image](https://user-images.githubusercontent.com/53124295/73410319-b23c5180-42b6-11ea-9ac5-402c5074289d.png)

This is my environment.yml

```
name: python-environment 
channels:
  - conda-forge
dependencies:
  - python=3.6.*
  - ipywidgets
  - ipympl
  - pythreejs
  - ipyleaflet
  - resonance
  - opty
  - pandas
  - pydy
  - numexpr
  - matplotlib
  - nodejs
  - scipy
  - seaborn
  - scikit-learn
  - scikit-image
  - sympy
  - cython
  - numba
  - bokeh
  - numpy
  - astropy
  - bqplot
  - nb_conda_kernels
  - statsmodels
  - patsy
  - requests
  - cloudpickle
  - dill
  - dask
  - sqlalchemy*
  - hdf5
  - h5py
  - vincent
  - beautifulsoup4
  - protobuf
  - xlrd
  - numba
  - gnuplot
  - ghostscript
  - rpy2
```

This is my postBuild

```
#!/bin/bash

# A way for mybinder.org to realize it should install the jupyter lab extension
# automatically. This file does not affect anything except when mybinder.org use
# it for this specifically.
jupyter labextension install @jupyter-widgets/jupyterlab-manager@^1.0.1
jupyter labextension install jupyterlab_bokeh@1.0.0 --no-build
jupyter labextension install jupyter-matplotlib@0.5.0
```

martinRenou:
FontAwesome seems to be missing on the page. ipywidgets and jupyter-matplotlib just assume FontAwesome is available (as Jupyter Notebook and JupyterLab provide it).

You can quickly fix it by adding font-awesome on your page 

```
<script src="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/5.12.1/js/all.min.js" integrity="sha256-MAgcygDRahs+F/Nk5Vz387whB4kSK9NXlDN3w58LLq0=" crossorigin="anonymous"></script>
```

[...]

PR #294, opened 2020-10-27
Add ipympl example documentation

celine168:
There is an error I'm running into from using the ipympl github repo as the Binder repo:
Regarding #185 
```
---------------------------------------------------------------------------
FileNotFoundError                         Traceback (most recent call last)
<ipython-input-1-38257e2684a4> in <module>
----> 1 get_ipython().run_line_magic('matplotlib', 'widget')
      2 import matplotlib.pyplot as plt
      3 fig, ax = plt.subplots()
      4 fig.canvas.layout.width = '7in'
      5 fig.canvas.layout.height= '5in'

/srv/conda/envs/notebook/lib/python3.7/site-packages/IPython/core/interactiveshell.py in run_line_magic(self, magic_name, line, _stack_depth)
   2305                 kwargs['local_ns'] = sys._getframe(stack_depth).f_locals
   2306             with self.builtin_trap:
-> 2307                 result = fn(*args, **kwargs)
   2308             return result
   2309 

</srv/conda/envs/notebook/lib/python3.7/site-packages/decorator.py:decorator-gen-108> in matplotlib(self, line)

/srv/conda/envs/notebook/lib/python3.7/site-packages/IPython/core/magic.py in <lambda>(f, *a, **k)
    185     # but it's overkill for just that one bit of state.
    186     def magic_deco(arg):
--> 187         call = lambda f, *a, **k: f(*a, **k)
    188 
    189         if callable(arg):

/srv/conda/envs/notebook/lib/python3.7/site-packages/IPython/core/magics/pylab.py in matplotlib(self, line)
     97             print("Available matplotlib backends: %s" % backends_list)
     98         else:
---> 99             gui, backend = self.shell.enable_matplotlib(args.gui.lower() if isinstance(args.gui, str) else args.gui)
    100             self._show_matplotlib_backend(args.gui, backend)
    101 

/srv/conda/envs/notebook/lib/python3.7/site-packages/IPython/core/interactiveshell.py in enable_matplotlib(self, gui)
   3405                 gui, backend = pt.find_gui_and_backend(self.pylab_gui_select)
   3406 
-> 3407         pt.activate_matplotlib(backend)
   3408         pt.configure_inline_support(self, backend)
   3409 

/srv/conda/envs/notebook/lib/python3.7/site-packages/IPython/core/pylabtools.py in activate_matplotlib(backend)
    314     # when this function runs.
    315     # So avoid needing matplotlib attribute-lookup to access pyplot.
--> 316     from matplotlib import pyplot as plt
    317 
    318     plt.switch_backend(backend)

/srv/conda/envs/notebook/lib/python3.7/site-packages/matplotlib/pyplot.py in <module>
   2280     dict.__setitem__(rcParams, "backend", rcsetup._auto_backend_sentinel)
   2281 # Set up the backend.
-> 2282 switch_backend(rcParams["backend"])
   2283 
   2284 # Just to be safe.  Interactive mode can be turned on without

/srv/conda/envs/notebook/lib/python3.7/site-packages/matplotlib/pyplot.py in switch_backend(newbackend)
    219         else "matplotlib.backends.backend_{}".format(newbackend.lower()))
    220 
--> 221     backend_mod = importlib.import_module(backend_name)
    222     Backend = type(
    223         "Backend", (matplotlib.backends._Backend,), vars(backend_mod))

/srv/conda/envs/notebook/lib/python3.7/importlib/__init__.py in import_module(name, package)
    125                 break
    126             level += 1
--> 127     return _bootstrap._gcd_import(name[level:], package, level)
    128 
    129 

~/ipympl/backend_nbagg.py in <module>
     25 
     26 here = os.path.dirname(__file__)
---> 27 with open(os.path.join(here, 'static', 'package.json')) as fid:
     28     js_semver = '^%s' % json.load(fid)['version']
     29 

FileNotFoundError: [Errno 2] No such file or directory: '/home/jovyan/ipympl/static/package.json'
```

### DIFF
--- a/docs/examples/ipympl_example.rst
+++ b/docs/examples/ipympl_example.rst
@@ -67,6 +67,7 @@ Press the "Activate" button below to connect to a Jupyter server:
 
 .. raw:: html
 
+   <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/4.7.0/css/font-awesome.css" integrity="sha512-5A8nwdMOWrSz20fDsjczgUidUBR8liPYU+WymTZP1lmY9G6Oc7HlZv156XqnsgNUzTyMefFTcsFH/tnJE/+xBg==" crossorigin="anonymous" />
    <script src="https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.4/require.min.js"></script>
 
    <button id="activateButton" style="width: 120px; height: 40px; font-size: 1.5em;">
@@ -92,6 +93,15 @@ Press the "Activate" button below to connect to a Jupyter server:
      }
    </script>
    <script src="https://unpkg.com/thebelab@latest/lib/index.js"></script>
+
+
+.. warning:: This cell is only required in the thebe documentation due to the way that ipympl's Binder container is setup. Do not add or execute it when working with ipympl yourself.
+
+.. raw:: html
+
+   <pre data-executable="true" data-language="python">
+   rm -r ipympl
+   </pre>
 
 2D plot
 -------
--- a/docs/examples/ipympl_example.rst
+++ b/docs/examples/ipympl_example.rst
@@ -1,0 +1,127 @@
+==============
+Ipympl Example
+==============
+
+Thebe can display output from ipympl_, which enables interactivity with matplotlib_.
+
+.. _ipympl: https://github.com/matplotlib/ipympl 
+.. _matplotlib: https://matplotlib.org/ 
+
+Setup
+=====
+
+Be sure to load require.js and Font Awesome 4 before any of your thebe activation code so that the
+matplotlib widgets can function. 
+
+.. code-block:: html
+
+   <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/4.7.0/css/font-awesome.css" integrity="sha512-5A8nwdMOWrSz20fDsjczgUidUBR8liPYU+WymTZP1lmY9G6Oc7HlZv156XqnsgNUzTyMefFTcsFH/tnJE/+xBg==" crossorigin="anonymous" />
+   <script src="https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.4/require.min.js"></script>
+
+
+Configure thebe and load it:
+
+.. code-block:: html
+
+   <script type="text/x-thebe-config">
+     {
+       requestKernel: true,
+       binderOptions: {
+         repo: "matplotlib/ipympl",
+         ref: "0.5.8",
+         repoProvider: "github",
+       },
+     }
+   </script>
+   <script src="https://unpkg.com/thebelab@latest/lib/index.js"></script>
+
+Create a button to activate thebe:
+
+.. code-block:: html
+
+   <button id="activateButton" style="width: 120px; height: 40px; font-size: 1.5em;">
+     Activate
+   </button>
+   <script>
+   var bootstrapThebe = function() {
+       thebelab.bootstrap();
+   }
+   document.querySelector("#activateButton").addEventListener('click', bootstrapThebe)
+   </script>
+
+Now add  between these HTML tags:
+
+.. code-block:: html
+
+   <pre data-executable="true" data-language="python">
+   %matplotlib widget
+
+   </pre>
+
+Examples
+========
+
+Using ipympl, you can display a variety of interactive plots.
+
+Press the "Activate" button below to connect to a Jupyter server:
+
+.. raw:: html
+
+   <script src="https://cdnjs.cloudflare.com/ajax/libs/require.js/2.3.4/require.min.js"></script>
+
+   <button id="activateButton" style="width: 120px; height: 40px; font-size: 1.5em;">
+     Activate
+   </button>
+   <script>
+   var bootstrapThebe = function() {
+       thebelab.bootstrap();
+   }
+   document.querySelector("#activateButton").addEventListener('click', bootstrapThebe)
+   </script>
+
+.. raw:: html
+
+   <script type="text/x-thebe-config">
+     {
+       requestKernel: true,
+       binderOptions: {
+         repo: "matplotlib/ipympl",
+         ref: "0.5.8",
+         repoProvider: "github",
+       },
+     }
+   </script>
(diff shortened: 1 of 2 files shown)


martinRenou:
@celine168 how did you install ipympl?

Even though I don't fully understand why you get this error, I've opened an issue on the ipympl repo. I think we can get rid of this line of code that fails.

celine168:
@martinRenou thank you for opening the issue on the ipympl repo! 

I am trying to use the Binder image spun up from the matplotlib/ipympl repository here:
```
     {
       requestKernel: true,
       binderOptions: {
         repo: "matplotlib/ipympl",
         ref: "0.5.8",
         repoProvider: "github",
       },
     }
   </script>
   <script src="https://unpkg.com/thebelab@latest/lib/index.js"></script>
```

celine168:
The issue is strange because when I try to spin up https://github.com/matplotlib/ipympl with tag 0.5.8 on Binder, the example notebooks seems to work fine.

moorepants:
@celine168 Can you check this again? Curious if the error is the same.

celine168:
Unfortunately it still doesn't seem to work. 
I noticed something new though: when creating a notebook in the root of the https://github.com/matplotlib/ipympl repository, I run into the error. When creating a notebook in the `examples` directory, the code executes fine. 

Part of the error is this:
```
~/ipympl/backend_nbagg.py in <module>
     25 
     26 here = os.path.dirname(__file__)
---> 27 with open(os.path.join(here, 'static', 'package.json')) as fid:
     28     js_semver = '^%s' % json.load(fid)['version']
     29 
```
I'm not sure where this file is executing but maybe it's dependent on where the current notebook is?

martinRenou:
How is ipympl installed?

moorepants:
ipympl is installed via docker2repo using this file https://github.com/matplotlib/ipympl/blob/master/environment.yml in the main repository. We are using the same mechanism as the "launch binder" button on the ipympl repo.

moorepants:
@celine168 can @martinRenou reproduce by launching the ipmpl repo on the libretexts binder?

martinRenou:
> I noticed something new though: when creating a notebook in the root of the https://github.com/matplotlib/ipympl repository, I run into the error.

I actually think that's the key to the error.

When running a Notebook at the root of the repository, `import ipympl` might import the **local** Python module, not the one that is installed, which is wrong.

When running a Notebook in another place, `import ipympl` will import the installed Python module (under site-packages, which contains the ipympl module with the package_data: package.json).

This makes me really want to get rid of the `package.json` file in the package_data, this way both approaches should work just fine.

martinRenou:
So as a conclusion I would say that there is nothing wrong with your installation, the only problem is the location of the file where you execute `import ipympl`.

moorepants:
We need ipympl to work regardless of which directory `import ipympl` is executed. I guess then, we just need a repository with only the environment.yml file. If we try to use the ipympl repository, thebe will be executed in the top level directory of that repo, so it is bound to pick up the ipympl package in that directory instead of an installed one.

@celine168 There are probably a few ways to "fix" this. For the thebe examples to always work, we need some way to launch the binder instance with only the `environment.yml`. Is that something that can be set in the thebe config? Maybe the file could be hosted in this repo and the proper config set to point to it?

martinRenou:
> We need ipympl to work regardless of which directory import ipympl is executed

I agree, in that sense, I would say ipympl is the problem, not Thebe. But you can indeed work around it by not having the ipympl repository around.

celine168:
> How is ipympl installed?

I'm actually running the repo via Binder, not locally. You can try it out here: https://mybinder.org/v2/gh/matplotlib/ipympl/stable
@martinRenou  Thank you so much for the clarification on the package.json file.

celine168:
@moorepants Would hosting an `environment.yml` file in the root of this repository to be used by repo2docker be confusing for other users? 

I like the idea, though I'm slightly worried that as we add more example pages and packages, that the packages for other pages might break? Maybe it's an unnecessary worry.

moorepants:
I think you could drop a `ipympl-env.yml` file in the docs directory in this repo. We could do that for each example that needs it. That file belongs in either this repo or the ipympl one. I'm not sure where else it should go. If in the ipympl repo we need to patch ipympl so that it can run from the directory root.

manics:
> Would hosting an environment.yml file in the root of this repository to be used by repo2docker be confusing for other users?

You can put the repo2docker file under a top-level `binder/` or `.binder/` directory instead of in the root of the repository: https://repo2docker.readthedocs.io/en/latest/usage.html#where-to-put-configuration-files

moorepants:
I don't think we want any environment.yml files in the root of this repo or in a `.binder/` as we need different repo2docker config files for each example in the docs. We'd like to put it adjacent to the specific sphinx rst file in `docs` and then tell binder to build with that file for that specific example page. Can we specify in the thebe config where exact the repo2docker config file is? (instead of relying on it being in these default locations?)

moorepants:
I don't see any thebe config vars that would let us do that. That makes it a bit difficult. Do we need to (or want to) create `github.com/executablebooks/ipympl-binder` for the sole purpose of hosting the single `environment.yml` file for a functioning ipympl binder? Not sure, as it will get out of hand if we have to do this for more example pages. It would be nice if ipympl hosted a functioning repository for ipympl to work with repo2docker instead (I suppose).

docs/examples/ipympl_example.rst
```diff
@@ -0,0 +1,127 @@
...
+       requestKernel: true,
+       binderOptions: {
+         repo: "matplotlib/ipympl",
+         ref: "0.5.8",
+         repoProvider: "github",
+       },
+     }
+   </script>
+   <script src="https://unpkg.com/thebelab@latest/lib/index.js"></script>
+
```
moorepants, line 95: I think you should just add this line here:

```
.. warning: This cell is only required in the thebe documentation due to the way that ipympl's binder container is setup. Do not add or execute it when working with ipympl yourself.

.. raw:: html
   <pre data-executable="true" data-language="python">
    rm -r ipympl
   </pre>

```

celine168: Just added this cell block in!

celine168:
Here's a preview of what the documentation currently looks like: 

image

moorepants:
Nice!

moorepants:
This LGTM, merging!

PR #295, opened 2020-10-27
Add bqplot documentation

celine168:
image

### DIFF
--- a/docs/examples/bqplot_example.rst
+++ b/docs/examples/bqplot_example.rst
@@ -60,6 +60,7 @@ Example
 
 .. raw:: html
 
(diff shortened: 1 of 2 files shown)


moorepants:
LGTM, merging as is.

choldgraf:
fantastic! It worked great for me as well 👍

moorepants:
I just tried it again and wasn't able to get the panzoom button to do anything. I clicked it, a blue edge appears around the button, and I'd expect that if I click, hold, and drag in the plot that it would pan, but nothing happens. Maybe I'm using the button wrong?

celine168:
Ah yes, you are right. My bad on that, I didn't specifically test panning/zooming on the plot. 

I think there might be an underlying issue with this.

celine168:
@moorepants @choldgraf would it be best practice for me to submit a PR reverting deleting the page before finding the fix?

choldgraf:
I'd also be fine with adding a note about zoom not working, and opening an issue to track this problem.

PR #297, opened 2020-10-27
Squash readOnly base-case tests to a single test

Miniland1333:
Reviewer indicated that the multiple `base-case` tests were confusing and did not improve testing capabilities. The three `base-case` test have been simplified into one test case of `noconfig`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -82,6 +82,7 @@ jobs:
       - name: "Run tests"
         run: |
           npm test
+          npm run test:jest
 
       # Action Repo: https://github.com/codecov/codecov-action
       - name: "Upload coverage to codecov"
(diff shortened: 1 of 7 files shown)


minrk:
Nice! Can you add `npm run test:jest` to the CI workflow so these tests get run?

Miniland1333:
I can do that, though once #282 is ported to Jest we might end up completely replacing the single Karma tests with the Jest tests.

Miniland1333:
I have added `npm run test:jest`, though again we will likely eventually modify `npm test` directly when swapping from Karma to Jest. Also, is Code Coverage something you would like to eventually enable? We don't have it configured but the tests are generating coverage.

minrk:
Enabling coverage reports would be great! Definitely would be nice to merge to just one test runner and remove karma.

PR #298, opened 2020-10-30
using local thebe in docs

choldgraf:
An attempt at getting the latest thebe to work in our docs. This does the following things:

- In `conf.py` (so, whenever docs are built with Sphinx), check if `_static/thebe` exists.
- If not, then run `npm build:prod` and copy the `lib/` folder to `_static/thebe`
- In the docs, always link to `_static/thebe/index.js` when we load thebe (even though the `code-block` examples will show loading from the CDN)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,5 @@ coverage
 webpack.stats.*
 docs/_build
 /.idea/
+
+docs/_static/thebe
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -164,6 +164,7 @@ import shutil as sh
 
 if not Path("_static/thebe").exists():
     print("Couldn't find local `thebe` build, building now...")
+    run("npm install".split(), cwd="..")
     run("npm run build:prod".split(), cwd="..")
     sh.copytree("../lib", "_static/thebe")
 else:
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -156,3 +156,15 @@ epub_exclude_files = ['search.html']
 
 # -- Linkcheck options ------------------
 linkcheck_anchors_ignore = ["/#!"]
+
+# -- Build the latest JS for local preview -----------------------------
+from subprocess import run
+from pathlib import Path
+import shutil as sh
+
+if not Path("_static/thebe").exists():
+    print("Couldn't find local `thebe` build, building now...")
+    run("npm run build:prod".split(), cwd="..")
+    sh.copytree("../lib", "_static/thebe")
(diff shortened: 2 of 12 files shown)


choldgraf:
arggg I cannot figure out how to get `node` to upgrade to 10.x so that we can build this package on RTD. @minrk have you ever had to update `node` on RTD before?

celine168:
Hi, I've tried to checkout this PR to check if I could find a fix.
I ended up getting an error like this, but I suspect that there is something wrong with my Python installation?
```
Running Sphinx v3.2.1
Couldn't find local `thebe` build for docs, building now...

Configuration error:
There is a programmable error in your configuration file:

Traceback (most recent call last):
  File "c:\users\celine\appdata\local\programs\python\python36\lib\site-packages\sphinx\config.py", line 319, in eval_config_file       
    exec(code, _globals)
  File "C:\Users\celine\college\libretexts\thebe\docs\conf.py", line 173, in <module>
  File "c:\users\celine\appdata\local\programs\python\python36\lib\subprocess.py", line 423, in run
    with Popen(*popenargs, **kwargs) as process:
    restore_signals, start_new_session)
  File "c:\users\celine\appdata\local\programs\python\python36\lib\subprocess.py", line 1017, in _execute_child
    startupinfo)
FileNotFoundError: [WinError 2] The system cannot find the file specified

make: *** [html] Error 2
```

stevejpurves:
@choldgraf did you get past the v1 issue? I think you were trying to update node on RTD, looking into it the default report for Ubuntu 18.04 use node v10.x so if it is not already on the build container (which it should be) then adding is to the `build/apt_pacakges` list should do it?

reference: https://docs.readthedocs.io/en/stable/config-file/v2.html#build

choldgraf:
@stevejpurves nope I couldn't figure this out, and then didn't have time to make another big push on it. If you like, feel free to pick up this PR (or to clobber it with a new PR). What's your readthedocs username? I can add you as an admin on that site so that you can play with config if necessary.

stevejpurves:
@choldgraf is am `stevejpurves` on readthedocs

I'll take a look at this, but I am also wondering if we need npm at all? could we not just fetch the package from `unpkg` itself using requests and unpack it into static ourselves... 🤔

choldgraf:
I am really not familiar with JS workflows so will defer to your and others intuition on this!

stevejpurves:
closing in favor of https://github.com/executablebooks/thebe/pull/472

PR #301, opened 2020-11-03
Docs: Add bqplot pan/zoom exception

celine168:
Add note about the pan/zoom button not working to the bqplot documentation.


minrk:
Thanks!

PR #345, opened 2021-01-06
🐛 FIX: Should request kernel if needed when restartAndRunAll

### LINKED ISSUE #341: Restart & Run All does nothing if kernel hasn't been requested
**Describe the bug**

If kernel hasn't been requested, `Restart and Run All` button will change output text to `waiting for kernel...` and hangs.

**To Reproduce**

1. Set `kernel: false` in configuration
2. Click on `Restart and Run All`

**Expected behavior**

Should request kernel first and then run all the cells.

**Possible Fix**

Removing the kernel check in [here](https://github.com/executablebooks/thebe/blob/master/src/thebelab.js#L282) should fix the problem since kernel is checked in `execute` and will automatic request if not found.

narrator0:
resolves #341 

### What is the problem?

When kernel is not requested yet, pressing `Restart and Run All` will hang in `Waiting for kernel` forever.

### What is expected?

Should go ahead and request a kernel and then run all.

### What is the fix?

Do not check if kernel exist in `restartAndRunAll()` since it will call `execute()` and that will do the job if a kernel is not there.

### DIFF
--- a/src/thebelab.js
+++ b/src/thebelab.js
@@ -279,7 +279,7 @@ function renderCell(element, options) {
       window.thebelab.cells.map((idx, { setOutputText }) => setOutputText());
     }
     restart().then((kernel) => {
-      if (!kernel || !window.thebelab) return kernel;
+      if (!window.thebelab) return kernel;
       // Note, the jquery map is overridden, and is in the opposite order of native JS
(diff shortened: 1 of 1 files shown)


choldgraf:
thanks very much, this looks reasonable to me!

PR #363, opened 2021-02-23
Ipycytoscape documentation repository --> 1.2.0

sandertyu:
The current repository `QuantStack/ipycytoscape 1.0.4` seems to be broken on the [example page](https://thebelab.readthedocs.io/en/latest/examples/ipycytoscape_example.html) for Thebe. There are errors in console, and the widget does not display. This PR simply bumps the Thebe config in this documentation to a newer version of this repository, `1.2.0`, which I have confirmed works on local HTML pages. This tag runs on jupyterlab3 rather than 2.


moorepants:
Thanks, merging!

PR #380, opened 2021-04-06
Make config example parseable

antoine-galataud:
Noticed a missing comma in configuration example. Just to save half a second for the next one who's going to copy/paste it.


moorepants:
Thanks!

PR #381, opened 2021-04-09
Updated changelog with two fixes.

moorepants:


choldgraf:
@moorepants do you think we should just make a release? Neither of those changes are breaking and seem to fix annoying UX problems

moorepants:
I'm making it right now. And forgot about this PR in the process. I still can't figure out #382.

moorepants:
The release worked, see: https://www.npmjs.com/package/thebe

moorepants:
Sorry it was a bit messy. I was confounded by these npm dependency errors/warnings.

Issue #422, opened 2021-07-19
Run cell busy/progress indicator


stevejpurves:
**the problem**

At the moment, after a kernel is available and connected, when we press [run] on a cell there is no visual feedback at all. The action buttons themselves stay enabled, and there is no other indication of busy'ness available. 

In some cases, execution is so fast that this is not noticeable but for other use cases for example the Leaflet and Cytoscape examples, is tasks a few seconds for the execution to complete.

**the desired solution**
This leads to impatience and multiple clicks. It's also not clear what the impact of multiple clicks are or if they are indeed registered. This leaves the user confused.

A simple improvement would be to add some form of busy indicator, to at least show the click has been registered.

**Describe alternatives you've considered**

We can also consider disabling the buttons once execution has been requested but it's not clear if edge cases are currently relying on that.  We would disable only the run button and have this re-enabled when on success, or failure, or kernel restart. I think this should be done in addition to a busy indicator if at all.

psychemedia:
In terms of UI, in classic notebook I use a cell status indicator extension that highlights cell execution in the margin:

![image](https://user-images.githubusercontent.com/82988/126900337-c72a9ae0-c73f-4608-b01a-ad024ee1eaaf.png)

Something a bit subtler showing a coloured vertical bar in the gutter might be enough of an indication.

choldgraf:
I believe @stevejpurves actually already tackled this one! @stevejpurves is this good to close?

stevejpurves:
ah yes, this is good to close as the PR was merged. It's a spinner style indicator rather than the linear progress, @psychemedia once that's out would be interested in your take on it.

PR #383, opened 2021-04-09
Try using node 15 on CI.

moorepants:
Fixes #382

Issue #382, opened 2021-04-09
`git diff --exit-code` is failing on master


moorepants:
**Describe the bug**
 
Seeing:

```
Run # If this fails, run `npm run fmt` and push the result

> thebe@0.6.0 fmt /home/runner/work/thebe/thebe
> prettier --trailing-comma=es5 --write *.js src/*.js examples/*.js test/*.js

babel.config.js 36ms
jest-puppeteer.config.js 8ms
jest.config.js 10ms
jest.e2e.config.js 17ms
jest.setup.js 4ms
jupyter-server.js 11ms
webpack.config.js 33ms
src/emptyshim.js 8ms
src/index.js 6ms
src/manager.js 28ms
src/output.js 42ms
src/services-shim.js 39ms
src/thebelab.js 215ms
examples/thebe_status_field.js 18ms
test/bootstrap.spec.js 12ms
test/helpers.js 5ms
test/render.spec.js 17ms
test/sanity.spec.js 4ms
test/thebe.properties.spec.js 4ms
diff --git a/package-lock.json b/package-lock.json
index f834949..3884625 100644
--- a/package-lock.json
+++ b/package-lock.json
@@ -3640,25 +3640,6 @@
         }
       }
     },
-    "@jupyterlab/codeeditor": {
-      "version": "3.0.0-rc.5",
-      "resolved": "https://registry.npmjs.org/@jupyterlab/codeeditor/-/codeeditor-3.0.0-rc.5.tgz",
-      "integrity": "sha512-eIY5J1KXs2E5J9IXhAM8TQFMqhe2N5TMheOgTyWq6J+9oPbSauxNvvGzrWn8Y9aBsDw5i2QamZRn82VnwlOn9A==",
-      "dev": true,
-      "requires": {
-        "@jupyterlab/coreutils": "^5.0.0-rc.5",
-        "@jupyterlab/nbformat": "^3.0.0-rc.5",
-        "@jupyterlab/observables": "^4.0.0-rc.5",
-        "@jupyterlab/translation": "^3.0.0-rc.5",
-        "@jupyterlab/ui-components": "^3.0.0-rc.5",
-        "@lumino/coreutils": "^1.5.3",
-        "@lumino/disposable": "^1.4.3",
-        "@lumino/dragdrop": "^1.6.4",
-        "@lumino/messaging": "^1.4.3",
-        "@lumino/signaling": "^1.4.3",
-        "@lumino/widgets": "^1.14.0"
-      }
-    },
[...]

PR #424, opened 2021-07-19
✨ ENHANCE: Add Feat/busy indicator

stevejpurves:
Added a busy indicator to cells which addresses issue #422 

This is achieved by adding a css spinner, a new element, and some jquery around the requestExecute promise.

Q5dl3GyGgx

This also works when restart and run all fires.

Reflecting busy status could be taken further as discussed in the issue, but probably in other ways too.


choldgraf (commented):
This looks great, thanks for the improvement :-) 

LGTM with a few thoughts:

- Will the spinner eat up a lot of CPU for users? I don't think it's a huge deal, but I've seen CSS spinning turn on laptop fans before :-)
- You deleted package-lock...that intentional?
- Can you fix the lint test?

moorepants:
We've had requests for this with LibreTexts users. A welcome addition!

moorepants:
This looks good to merge from me, other than the deleted json file and the test. It seems the failure has something to do with the deleted file.

stevejpurves:
Re: the package-lock.json

It was an intentional update but after looking deeper into the test failure I'd like to back out of that for now. So I am intending to revert.

However, I think there is a problem with the git workflow itself. In the lint step, as well as running `npm run fmt` we also do a `git diff` on the local checked-out repo.

That is what is failing as, in that step we also do an `npm install`. `npm install` will install fresh packages and create a new package-lock.json file, this newly updated file is different from the checked-in version because we have loosely versioned packages in our package.json. i.e. we are using the `^` on package versions. This is fine as long as we use `npm ci` instead of `npm install` in the github action.

I've changed the workflow to use `npm ci`

We can update the `package-lock.json` further when we bring in dependabot alerts.

moorepants:
Yes, this git-diff thing has caused me headaches in the past too. I didn't understand why it is there.

moorepants:
Here is the issue I discussed the diff: https://github.com/executablebooks/thebe/issues/382

stevejpurves:
OK, this is still failing for me because I am running with npm v6.x locally. It seems like npm v7.x is now the minimum supported version (looking at @moorepants's comments here https://github.com/executablebooks/thebe/pull/383)

We need to update engines in package.json to prevent this problem. Currently we have:
```
  "engines": {
    "node": ">=6",
    "npm": ">=4"
  },
```

npm can go to >=7, but should we also move node to >=15? v15 was the first to ship npm v7 by default.

stevejpurves:
@choldgraf I think these spinners are ok. They are a single disc with a single css transform, so the most minimal possible. tested locally and left them running and didn't see any blip in cpu. Not exactly a performance test but they seem ok.

moorepants:
I'm fine with the lower node versions if you can get things to work. My "fix" with npm 15 was because that's all I could figure out. There did seem to be a mismatch in what node/npm versions were install locall vs the CI, but part of that is me not really understanding all this in javascript dev.

stevejpurves:
ok, it's all green now.

So we now have the engines and min requirements in the docs pointing to the same version as in use in CI. These are also the most recent versions so that is a good thing imo.

Note this won't force users to upgrade their node environments, they can still use earlier ones but get warning messages during install but it does mean there is a clear minimum requirement that developers should meet.

docs/contribute.md
```diff
@@ -36,9 +44,11 @@ Next, clone the repository and set up your `npm` environment for this repo:
 ```bash
 git clone https://github.com/executablebooks/thebe
 cd thebe
-npm install
+npm ci
 ```
 
+NOTE: using `npm ci` rather than `npm install` will ensure that you install the latest tested dependencies, and will not make any unintentional local upgrades
```
choldgraf, line 50:
````suggestion
```{note}
Using `npm ci` rather than `npm install` will ensure that you install the latest tested dependencies, and will not make any unintentional local upgrades
```
````

File: docs/contribute.md [hunk 2]
```diff
@@ -109,20 +119,21 @@ e2e style tests are achieved using [Puppeteer](https://github.com/puppeteer/pupp
...
-    );
-  });
- ```
- - Assert on initial page state
- - Invoke UI actions to trigger behavior
- - assert on final state
+- creating a test html page that load and uses thebe, placing this in the `e2e/fixtures/HTML` folder
+- load the fixture page at the start of your test
 
+```
```
choldgraf, line 125:
```suggestion
```javascript
```

choldgraf (approved):
This looks good to me - thanks for the improvement @stevejpurves. I'll let somebody else with more JS knowledge than myself push the green button, or I'll press it in a day or two if nobody else chimes in :-)

choldgraf:
thanks @stevejpurves for the improvement!

moorepants:
The lint test failed on master. We need to dig into that a bit more.

PR #454, opened 2021-08-16
:recycle: REFACTOR: splitting out responsibilities #1

stevejpurves:
This PR is the first step in proposed internal refactoring - previous to this all the code lived in one file, `thebelab.js` making reading and reasoning about the code harder than it needed to be. this is a first step in splitting out that code into separate files, so that in a subsequent PR we can start better-defining interfaces, where appropriate use objects to encapsulate the different responsibilities and start putting tests in place around them.

Code has been split out into the following files/areas:

 - options
 - kernels
 - render
 - utils

`thebelab` still contains the bootstrap function and pulls together the top-level namespace object.

The intention is that the exported external API does not change, and functions have been exported from modules accordingly.

**Testing**
 - tests pass
 - docs build and examples in there are working
 - local and binder based development environments are working


moorepants:
@choldgraf Can we give @stevejpurves commit rights for this repo?

stevejpurves:
I've moved events out into a separate file (module), which addresses the issue I raised above

choldgraf:
I believe @stevejpurves should have commit rights now. LMK if that's not the case @stevejpurves 

@moorepants it would be useful if you and your team could confirm that this refactor does in fact make the code more sensible? I believe you have worked with the codebase much moreso than I :-)

moorepants:
We can have a look.

stevejpurves:
It would be great if you could @moorepants 👍 also I see this as a stepping stone and not the final state.

TimStewartJ:
Having worked with the code base a bit recently, I think this would be a huge help! Continuously scrolling through the huge `thebelab.js` file got exhausting quickly. Its separation in this PR makes much more sense.

Also, +1 on the testing, I've tried local and binder development environments as well.

moorepants:
I read through this. Looks good. As long as this is true:

> The intention is that the exported external API does not change, and functions have been exported from modules accordingly.

then I'm fine with this change and it can be merged.

rowanc1:
(@stevejpurves is away this week, so he probably won't be merging this on his own!)

moorepants:
I'll merge.

Issue #343, opened 2021-01-02
Consider allowing multiple kernels in the same document.


akhmerov:
**Is your feature request related to a problem? Please describe.**

Historically jupyter-sphinx [implemented](https://jupyter-sphinx.readthedocs.io/en/latest/#controlling-the-execution-environment) including the inputs and outputs of *multiple* notebooks within the same document. This was done before we had thebe integration. Thebe, on the other hand, assumes a correspondence 1 document = 1 kernel.

I would even consider removing this feature from jupyter-sphinx, however [searching](https://github.com/search?p=2&q=extension%3Arst+%22jupyter-kernel%22&ref=simplesearch&type=Code) through github revealed multiple uses out in the wild, including one by @moorepants in pydy and one in kwant where I'm involved. These uses fall into two categories:

- Section of a document that are conceptually isolated, e.g. different topics or different exercises.
- Parts to be executed in different languages. This makes sense in a library that provides API in multiple languages. Can even be combined with tabs... :thinking:.

Both uses seem reasonable.

**Describe the solution you'd like**

I therefore propose to support running multiple kernels within the document. This may be ruled as being out of scope of thebe, but I think this warrants consideration. One possible implementation would be to label each thebe DOM element with the kernel id to which it belongs.

**Describe alternatives you've considered**

Not supporting multiple kernels in a single document :)

choldgraf:
It seems like a reasonable feature to have, though it breaks with Jupyter's "one kernel per document" approach, and we generally default to "doing things like Jupyter" unless there's a clear reason to depart.

I think a big thing we should consider is extra maintenance and complexity burden. Thebe is operating with basically 0 resources right now, so whatever solution we come up with needs to be fairly simple and straightforward, and cannot involve a major change in the code.

akhmerov:
I completely agree with the maintenance and engineering aspect. As I wrote in the description, I even considered disabling the feature completely until I saw that there are uses out in the wild. Labeling as "help wanted" based on that.

joergbrech:
BTW, we have another use case, where we want to write a "Modeling and Simulation" jupyter-book for engineers. Currently the book uses a Matlab/Octave kernel and that is the main language taught in the course. But the language itself is not the main focus of the course. In the future, we would like to add Julia and Python as well and ideally have all the interactive code blocks multilingual. I was thinking of something along the lines of interactive [tabbed content blocks](https://jupyterbook.org/content/content-blocks.html#tabbed-content).

[...]

Issue #271, opened 2020-10-17
Kernel Status


rowanc1:
Talking with @choldgraf on some improvements now that we are talking about what to do when a user restarts the kernel or if it is connected or not. I think that there may be some room to add a bit more state management that can be exposed in a flexible way to the UI.

Thinking of an object that powers a UI similar to:

![image](https://user-images.githubusercontent.com/913249/96353658-7f77fa80-108b-11eb-9635-9bf1ae3e5e6f.png)

Does something like this already exist on the session object?

See also: #38

choldgraf:
That sounds just like what I had in mind as well 👍

minrk:
Kernel status is available as [kernel.status](https://jupyterlab.github.io/jupyterlab/modules/_services_src_index_._services_src_kernel_kernel_.ikernelconnection.html#status) and there is a [kernel.statusChanged](https://jupyterlab.github.io/jupyterlab/modules/_services_src_index_._services_src_kernel_kernel_.ikernelconnection.html#statuschanged) signal we can use to trigger events when the kernel status is changed.

stevejpurves:
noticing that `sphinx-thebe` has a nice kernel launch indicator already, which could be a good pattern to start with in terms of adding a kernel status widget?

![image](https://user-images.githubusercontent.com/1473646/128700269-c7eed129-2cf9-41b7-837b-fb6c34d0c695.png)

stevejpurves:
also there is a status_field example, with js and css file in the `examples` folder. I think this could be brought into the main library and improved from there, adding some information into the docs too.

PR #459, opened 2021-08-16
Kernel Status

stevejpurves:
responding to issue #271 to make first-class support for a Status Widget. 

Previously, users had to rely on functions in the examples folder in order to add status fields to their page. In circumstances where a generic status display will fit, this is now provided within thebelab.

The widget is minimally styled and styles can be easily overridden.

image

This could also be extended to help manage the multiple kernel case requested in #343 and to give more information back on the kernel type.

TODO:

 - [x] add plain HTML example
 - [x] add examples in the main docs
 - [x] add some tests around KernelStatus
 - [x] fix the readthedocs build
 - [x] added a built in active button
 - [x] updated examples to show activate button and `launch` button as separate examples. The launch button showing dynamic library loading
 - [x] moved all library initialization into  the DOMContentsLoaded event handler
 - [x] added User Interface page to docs


stevejpurves:
I feel like this is ready to come in but it will be a pain to test locally in terms of the docs build ahead of either #472 or alternative coming in.

The purpose of this was to really,  get this common functionality inside the library instead of relying on consumers implementing it from the examples.

I think this meets that and I have not gone as far as bringing in the `sphinx-thebe` button, not yet anyways. I think instead this gives a minimum easy `activate` buttons that works ok with the status widget.

docs/ui.rst
```diff
@@ -0,0 +1,94 @@
+==============
+User Interface
+==============
+
+Thebe includes support for some built-in user interface elements.
+
+The aim is to provide simple elements that can be added to a page a styled accordingly by the consumer without neeing to [re]write coont code.
```
choldgraf, line 7:
```suggestion
The aim is to provide simple elements that can be added to a page and styled accordingly by the consumer without needing to [re]write code.
```

docs/ui.rst
```diff
@@ -0,0 +1,94 @@
...
+
+- Activate Button
+- Kernel Status Widget
+
+Activate Button
+===============
+
+Configure this UI element as follows.
+
+Add a :code:`div` element to the page in the desired location.
```
choldgraf, line 20:
```suggestion
Add a ``div`` element to the page in the desired location.
```

I think double backticks does the same thing in rST (we should really convert this to MyST one day 😬 )

docs/ui.rst
```diff
@@ -0,0 +1,94 @@
...
+
+Built in UI elements are:
+
+- Activate Button
+- Kernel Status Widget
+
+Activate Button
+===============
+
+Configure this UI element as follows.
```
choldgraf, line 18:
```suggestion
When this button is pressed, it will run the `thebelab.bootstrap` function on the page.
Configure this UI element as follows.
```

docs/ui.rst
```diff
@@ -0,0 +1,94 @@
...
+.. code-block:: html
+
+  <div class="thebe-activate" />
+
+Then set the following option to :code:`mountActivateWidget:true` in the Thebe config object
+
+Status Widget
+=============
+
+Configure this UI element as follows.
```
choldgraf, line 31:
```suggestion
This element will display the status of the kernel, it will update when the kernel sends new messages or when code is executed.
Configure this UI element as follows.
```

docs/ui.rst
```diff
@@ -0,0 +1,94 @@
...
+   %matplotlib inline
+   import numpy as np
+   import matplotlib.pyplot as plt
+   x = np.linspace(0,10)
+   plt.plot(x, np.sin(x))
+   plt.plot(x, np.cos(x))
+   </pre>
+
+
+Note: If you are looking to load Thebe dynamically, check the custom launch button in the example `here. <https://github.com/executablebooks/thebe/blob/feat/kernel-status/examples/demo-launch-button.html>`_
```
choldgraf, line 94:
```suggestion
Note: If you are looking to load Thebe dynamically, check the custom launch button `in the example here. <https://github.com/executablebooks/thebe/blob/feat/kernel-status/examples/demo-launch-button.html>`_
```

I find that it's easier to linkify multiple words so people can see the link more easily

src/options.js
```diff
@@ -27,8 +27,8 @@ const _defaultOptions = {
   stripPrompts: false,
   requestKernel: false,
   predefinedOutput: true,
-  mountStatusWidget: true,
-  mountActivateWidget: true,
+  mountStatusWidget: false,
```
choldgraf, line 30: any reason not to make these true by default? If a page doesn't have the `thebe-activate` etc class, then this would just not have any effect right?

choldgraf (commented):
A few quick thoughts from me! This looks really nice, and I appreciate the documentation as well.

For trying this out, is the best way still to clone locally and set up a local dev environment?

src/options.js
stevejpurves: yes, it's a no-op when they are not on the page. :+1: I'll set them to true

docs/ui.rst
stevejpurves: i like it 👍

choldgraf (approved):
This looks good to me...I can't review the JS in depth but it seems reasonable. I'd be +1 on merging this in and seeing how it feels in practice and making iterative improvements from there. WDYT @stevejpurves ?

stevejpurves:
planning on bringing this in tomorrow then - @moorepants @TimStewartJ your build doesn't have anything relying on the `thebe-status` or (new so unlikely) `thebe-activate` CSS classes?

moorepants:
No, we're good. We also don't run the tip of master in our plugin, so we can check these things on new releases. This is our code that uses thebe, for reference: https://github.com/LibreTexts/ckeditor-binder-plugin

choldgraf:
Cool - as long as this doesn't disrupt @moorepants (or any other communities using the tooling) then I'm +1 giving this a whirl!

choldgraf:
🚀

Issue #285, opened 2020-10-20
Use the latest `thebe.js` in our documentation


choldgraf:
I think that it would be best for us to use the latest `thebe.js` package in our documentation, so that we can both test it locally in a more "realistic" setup, and so that our docs can use features that may be available only in the "latest" release.

I think that the simplest way to do this would be via the `sphinx-thebe` package. I could modify that package to load `thebe.js` from a *local* path instead of from the CDN. Then we'd just need to build `thebe.js` as a single, local file and include it with our documentation in `_static`.

The only challenge is that I have no idea how to build `thebe.js` locally in a way that could be linked in a local SSG. Does anyone have guidance on how / the best way to do this?

moorepants:
Agreed, the versions of thebe used in the various docs versions should match. For example, if you want to write some explanation and demonstration about "Restart and Run All" you can't because it doesn't show up!

joelachance:
All of this seems sound to me-- I don't know about the Sphinx side of things, but we could write a custom git hook to run a script which would generate a `thebe.js` file inside the `docs/_static/`.  Of course there's other solutions, but that's the first one that comes to mind.

Does this sound like a viable solution, @choldgraf?

moorepants:
I'm not sure a git hook solves the issue. If you build the docs locally you'd like the version of thebe at that commit to be used to build the docs.

joelachance:
A git hook would merely run the script which would generate the build. It's a bonus, but not needed. The script could be an npm script living in package.json or something custom, but to your earlier point (a good one!), the script could generate versioned`thebe.js` builds, which could still be useful locally if we need to patch an older version (I don't know if we ever do this).
The 'build' step is pretty simple since we're already building Thebe, but the 'link in a local SSG' part I'm unsure about.

moorepants:
I don't know enough about modern JS tech, but from the python side most sphinx conf.py file simply append the parent directory to sys.path. This ensures your docs are built with local version of the python lib. Python libs don't have to be built though. If thebe has to be built before it can be used, then I guess adding that build step to the sphinx make file would be needed. Sorry if you already know all this...

choldgraf:
I feel like there are two options (both of which are fine with me):

[...]

PR #469, opened 2021-08-31
[DOC] use the latest code in builds of the docs via GHA build

stevejpurves:
This PR aims to address #285 and change the default docs build to always use the version of `thebe` represented by the latest code in the repo. Previously the docs would load the latest published version from `unpkg`.

TODOS:

 - [x] add makefile commands to build the docs locally
 - [ ] update `readthedocs` builds as needed and ensure that these build with the latest local `thebe`

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,4 +1,4 @@
-name: Docs
+name: build latest as artifact
 on:
   - push
   - pull_request
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,5 +1,7 @@
 name: Docs
-on: [push, release]
+on:
+  - push
+  - pull_request
 
 jobs:
   thebe-build:
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -5,7 +5,7 @@ on:
 
 jobs:
   thebe-build:
-    name: "Buil the current version of thebe for ue in the docs"
+    name: "Build the current version of thebe for use in the docs"
     runs-on: ubuntu-latest
     steps:
       # Action Repo: https://github.com/actions/checkout
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,48 @@
+name: Docs
+on: [push, release]
+
+jobs:
+  thebe-build:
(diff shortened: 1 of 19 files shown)


stevejpurves:
@choldgraf I started afresh here as I was concentrating on the local build, but intend to pull over changes from https://github.com/executablebooks/thebe/pull/298 as needed to close this out.

choldgraf:
Excellent! This will make it much easier to iterate. Is this ready for a review or do you wanna work on stuff further? I'd like to get this in ASAP

stevejpurves:
@choldgraf I need to get the `readthedocs` task done but will prioritize that

stevejpurves:
ok. Seems like there is no way to upgrade the node version on RTD even with the new `build.apt_packages` options.
This is a known issue and seems that it. is getting some attention here:

https://github.com/readthedocs/readthedocs-docker-images/issues/107

 As node is a dev dependency here, and I think we can use older versions ok. I've moved the engine requirement back to 8 and above to get this building until RTD updates their build system. I'm also leaving the current option file in place for now.

stevejpurves:
ok, so pausing this on until RTD folks update their docker image and options :(

stevejpurves:
Note: another way forward would be to use a github action to produce a local build on push, storing this as an artifact which the RTD build then consumes.

There is some good reference material in this repo https://github.com/dfm/rtds-action 

Taking over the thebe build and triggering the RTD build from this repo will give us full control over the JS build process. This seems like a good way to go! and one we can action asap.

choldgraf:
@stevejpurves I'm +1 on using GitHub actions if it means that we get control over the JS environment that is needed to make progress here. I think that's a good enough reason to use GHA over readthedocs for now. If so, we should open an issue about switching back to RTD once the issue you've linked is resolved, and then switch the thebe docs to use GHA instead of RTD.

stevejpurves:
@choldgraf great, let's try that. I'll need readthedocs access to setup the integration & have sent a message on EB slack 👍

stevejpurves:
Note: this will hopefully be superseded by https://github.com/executablebooks/thebe/pull/472

stevejpurves:
closing in favor of https://github.com/executablebooks/thebe/pull/472

Issue #464, opened 2021-08-24
Thebe loads CSS that clobbers in-page CSS


choldgraf:
### Describe the problem

When `thebe` is loaded, it now loads CSS that clobbers a ton of in-page CSS. For example, see the word "example" in the thebe docs here:

https://thebe.readthedocs.io/en/latest/#examples

![image](https://user-images.githubusercontent.com/1839645/130689627-304cb399-7f24-488b-9739-887494f27fb3.png)

In this case, all of the paragraph styling and the styling of the second instance of "examples" is gone.

If you look at the CSS that corresponds to those elements, we are loading a `base.css` file a bunch of times, that is coming from the JupyterLab UI components:

![image](https://user-images.githubusercontent.com/1839645/130689864-7335a8e0-a422-4105-afd6-37f29c848352.png)

So it seems like we are loading some JupyterLab CSS that is clobbering the other CSS on the page. We should restrict this to thebe-activated code cells somehow.

### Link to your repository or website

https://thebe.readthedocs.io/en/latest/#examples

### Steps to reproduce

- Load the latest `thebe` on any page with paragraph-level formatting
- Note that the formatting is clobbered

rowanc1:
I think my guess would be to look here:
https://github.com/executablebooks/thebe/blob/b3ddca5389ed80bb6add801259776577bb8ffb99/src/thebelab.js#L16

choldgraf:
Perhaps this is something that @moorepants or someone in his group has noticed as well? This seems like a pretty major change in end behavior

moorepants:
@TimStewartJ can you look into css behavior change with the new juplyterlab code?

moorepants:
> Perhaps this is something that @moorepants or someone in his group has noticed as well?

No we haven't seen these effects in our use cases.

stevejpurves:
ok so recreated this locally, using a local build and this line is the problem: https://github.com/executablebooks/thebe/blob/master/src/thebelab.js#L18

```
import "@jupyterlab/rendermime/style/index.css";
```

removing this no longer loads the css responsible for the bug. So at least this is a style file that we are directly loading.


This import was included in this commit: https://github.com/executablebooks/thebe/commit/1a8cb85c8b3eb903af511e7e22fcf696bf9bffaa
 with the message `add css with standard jupyter ansi colours to make output colourful`

[...]

PR #471, opened 2021-09-06
[FIX] jupyterlab css clobbers downstream css

stevejpurves:
This is a potential fix for issue #464.

Here I have avoided importing the `@jupyterlab/rendermime/style/index.css` file directly and instead import the dependencies listed in that index file, while omitting the files that eventually load the `@jupyterlab/ui-components/style/index.css` file.

I am not sure that we need any styles from `@jupyterlab/ui-components` at all but have made a local file `patch.css` that essentially recreates the `@jupyterlab/ui-components/style/index.css` but without the problematic rules on high-level selector. I'm not sure if we need this or not but wanted to align as closely as possible without including the problematic rules.

The sum of this is that we can load most of the css that was loaded previously. When loading the `ui-components` css and other css that is actually dependent on the rules that I have now omitted, there may be issues that need to be resolved.

So far I have tested with all the examples in the docs and they appear to all render correctly.

I'm interested in what other testing we could do here to ensure that this fixes the issue without causing further downstream issues? cc @minrk @choldgraf 

@moorepants @TimStewartJ  I was thinking that testing for any impact on the LibreTexts build would be a good further validation for starters?


src/patch.css
```diff
@@ -0,0 +1,93 @@
+/*-----------------------------------------------------------------------------
+| Copied from : https://raw.githubusercontent.com/jupyterlab/jupyterlab/master/packages/ui-components/style/base.css
+| with high level css selectors removed
+|
+| Note: this is highly fragile and may change each release of jupyterlab
```
choldgraf, line 5:
```suggestion
| Note: this is highly fragile and may change each release of jupyterlab
| See this issue for some context: https://github.com/executablebooks/thebe/issues/464
```

choldgraf (commented):
I think that this looks good to me. My assumption is that this won't make a huge difference because (I think) we don't re-use a ton of the lab UI components anyway.

I'm +1 as long as @moorepants confirms that this does not disrupt libretext builds in any way!

moorepants:
@TimStewartJ could you test this in the ckeditor plugin with our examples? We can check on libretexts after it is released.

TimStewartJ:
It appears to be working! I've gone ahead and made a quick patch to ckeditor-binder-plugin that uses this version that can be seen here: https://query.libretexts.org/Sandboxes/jupyterteam_at_ucdavis.edu.

moorepants:
Great, thanks Tim!

src/patch.css
```diff
@@ -0,0 +1,94 @@
+/*-----------------------------------------------------------------------------
```
moorepants, line 1: I believe we need to include JupyterLab's license in this repo if we copy code from there.

stevejpurves: yes, you're right. I really wonder if we need this patch though? I think I'll do some more testing with it removed, and if it stays we'll bring the in license.

stevejpurves:
I've gone ahead and removed the `patch.css` file after testing locally with the examples we have. I see no negative impact from removing it.

@moorepants @TimStewartJ I suggest that we bring this in but @TimStewartJ if you wanted to give it another quick double check, on your code that might be good. Sorry about the churn!

TimStewartJ:
Yep, sure thing. I've deployed this new patch to the same place and tested it. Everything seems to work as intended still.

stevejpurves:
great @TimStewartJ !

PR #472, opened 2021-09-07
[MNT,DOC] use latest code in builds of the docs with RTD

stevejpurves:
This PR aims to address #285 and change the default docs build to always use the version of thebe represented by the latest code in the repo. Previously the docs would load the latest published version from unpkg.

TODOS:

 - [x] add makefile commands to build the docs locally, and place a copy in `docs/_static`
 - [x] update docs to use the local build
 - [ ]  update readthedocs builds as needed and ensure that these build with the latest local thebe


humitos:
hi @stevejpurves! I'm testing this PR but I'm getting this error:

```
repos/thebe/docs  pr/472 ✗                                                                                                             2m ⚑  
▶ make html
rm -rf _static/lib
cd ..; yarn install --frozen-lockfile; yarn build
yarn install v1.22.11
[1/5] Validating package.json...
[2/5] Resolving packages...
error Your lockfile needs to be updated, but yarn was run with `--frozen-lockfile`.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
yarn run v1.22.11
$ webpack --mode development --progress
/bin/sh: line 1: webpack: command not found
error Command failed with exit code 127.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
make: *** [Makefile:43: js] Error 127
Time: 0h:00m:03s
```

it seems that `yarn` is not installing the dependencies and failing. I don't have too much experience with `yarn` and I'm not sure if it's possible to remove `--frozen-lockfile` argument.

I'm facing the same problem when testing this inside Read the Docs.

humitos:
For the record, I was able to make this PR work with the new Read the Docs Docker images and config file (see https://github.com/readthedocs/readthedocs.org/pull/8453, https://github.com/readthedocs/readthedocs.org/pull/8478, and https://github.com/readthedocs/readthedocs-docker-images/issues/107) with the following changes.

### Read the Docs configuration file

```yaml
# readthedocs.yml
version: 2

python:
  install:
    - requirements: docs/doc-requirements.txt

build:
  os: "ubuntu20"
  languages:
    python: "3.9"
    nodejs: "14"
```

Note the usage of `build.os` and `build.languages` (these name may change, we are still discussing them); but this is going to be the way to install nodejs, for example.

### Sphinx configuration file

```python
# docs/conf.py
path_root = Path(__file__).parent.parent
os.environ['PATH'] = f'{path_root}/node_modules/.bin/:' + os.environ["PATH"]

run(["npm", "install", "yarn", "jsdoc"], cwd=path_root)
run(["yarn", "--version"], cwd=path_root)
run(["yarn", "install"], cwd=path_root)
run(["yarn", "build",], cwd=path_root)
sh.copytree(f"{path_root}/lib", "_static/lib")
```

Highlights:

* I'm adding local `node_modules/.bin` to the PATH environment variable
* I removed the `--frozen-lockfile`
* I'm installing `jsdoc` that's required (not available by default in new Docker images)
* There is no need to call `node_modules/yarn/bin/yarn`, just `yarn` works

I'll let you know when we deploy these changes to production in beta so you can start testing this and giving us some feedback 😍

humitos:
> do you have a rough ETA on those changes making it to prod/beta?

Not yet. We are currently working to make this happen.

humitos:
Hi! We already deployed these changes. You should be able to use `build.os` and `build.tools` as I mentioned in https://github.com/executablebooks/thebe/pull/472#issuecomment-921065673. Please, let me know if it works for you! 💪🏼

choldgraf:
That's great news @humitos !!

@stevejpurves just lemme know if you'd like me to take a look or provide feedback on something!

stevejpurves:
ok, RTD build is green now. Guess I am surprised by `yarn install` seemingly cleaning out the `node_modules` directory and making `yarn`, `jsdoc` subsequently unavailable. With or without the `frozen-lockfile` option. I can't reproduce this locally, but its working on RTD.

Going to do a little more to try and make local dev cycle, nicer while keeping RTD up

stevejpurves:
This is ready for review and [squash 🙏 ] merge if it looks ok.

 - We're using the latest docker images from RTDs now for the build - thanks to @humitos & team. This went well, except for some unexpected `yarn` behavior [not what I expected anyways]. This now means that the RTD build is not using a `frozen-lockfile`, but I think that is acceptable.
  - I've removed the `prepare` npm script and replaced that with `prePublishOnly`. This was to avoid a double local build when running `make html`. We're expecting this not to affect the npm-publish step next time we cut a release but it's one to watch.
  - updated ci tests to build the latest lib explicitly, as `yarn install` no longer builds (because no `prepare` step)
  - HTML based examples now use local builds over latest release
  
  Hopefully, this doesn't negatively affect anyone's workflow.

docs/conf.py
```diff
@@ -156,3 +167,38 @@
...
+
+
+print('*************************** RUNNING CONF.PY ***************************')
+
+# -- Run JS Build if needed ------------------------
+# Note: this will be a one off run on RTD but may be run mulitple times locally
+# during development and testing
+path_root = Path(__file__).parent.parent
+
+if os.environ.get('READ_THE_DOCS'):
```
choldgraf, line 184: What's the reason for having two different workflows on RTD vs. locally? I'm wondering if it'd be possible to just have one check for whether the `_static/lib` folder exists rather than making it build environment-specific 🤔

choldgraf (commented):
This in general looks pretty nice to me! Just a few comments about the build process and the docs, happy to take another look or iterate on comments! Thanks @stevejpurves for doing this work :-)

docs/conf.py
stevejpurves: That's a good point @choldgraf. 

I'm seeing the RTD process as separate mostly due to the unique setup it needs i.e. because of the local installs of `yarn` and `jsdoc` which are typically going to be global installs in someone's local development environment. We don't want two versions of these kicking around locally, or to add them to the developer's path.

I have separated that environment setup from the build commands, where the latter are now a run based on a check for the `_static/lib` in both environments as you suggested.

Note in the course of this I've picked up on some more changes that were needed to properly support the "use local build" both locally and on RTD. see changes to `Makefile` and `webpack.config.js`.

package.json
```diff
@@ -7,10 +7,11 @@
     "build": "webpack --mode development --progress",
     "build:prod": "webpack --mode production",
     "build:watch": "webpack --mode development --progress --color --watch",
-    "prepare": "yarn run build:prod",
+    "prePublishOnly": "NODE_PREPUBLISH=true yarn run build:prod",
```
stevejpurves, line 10: Added environment var to identify `npm publish` action

package.json
```diff
@@ -7,10 +7,11 @@
...
     "fmt": "prettier --trailing-comma=es5 --write *.js src/*.js examples/**/*.js test/*.js",
     "serve": "http-server -c-1 -a 127.0.0.1 -o development/binder.html",
     "serve:local": "http-server -c-1 -a 127.0.0.1 -o development/local.html",
+    "serve:examples": "node setupExamples.cjs; http-server -c-1 -a 127.0.0.1 -o examples/index.html",
```
stevejpurves, line 14: added a serve command to make html examples useful in local development

docs/examples/matplotlib_interact_example.rst
```diff
@@ -67,7 +67,7 @@ Example
        },
      }
    </script>
-   <script src="https://unpkg.com/thebe@latest/lib/index.js"></script>
+   <script src="../_static/lib/index.js"></script>
```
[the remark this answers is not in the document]
stevejpurves: It would simplify that path, as we would be loading from `../lib/index.js` as per the other pure HTML examples that we are copying.

In this case this example needs sphinx to compile it to HTML first -- I'm not clear on how or when to move the result of that build into the HTML though?

choldgraf:
sounds good - just a quick thought, could we deprecate one of the build processes if it is somewhat redundant? E.g., could we:

- put the `examples/` folder inside of `docs/`
- serve that folder as a static asset with our docs (so HTML pages are just direct copied through)
- have them re-use the same `thebe` bundle that is packaged for use with `docs/`

Would that simplify?

docs/conf.py
```diff
@@ -156,3 +167,38 @@
...
+    node_modules_bin = f'{path_root}/node_modules/.bin/';
+    os.environ['PATH'] = f'{node_modules_bin}:' + os.environ["PATH"]
+    run(["npm", "install", "yarn", "jsdoc"], cwd=path_root)
```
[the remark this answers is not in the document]
choldgraf: just as a point of info - nox supports conda environments so you can use it with JS-style build systems, which is pretty nice. But I don't think it needs to block this! For example: https://github.com/pydata/pydata-sphinx-theme/blob/master/noxfile.py

docs/contribute.md
```diff
@@ -128,7 +128,21 @@ to run auto-formatting prior to each commit.
 
 ## Testing Thebe
 
-You can test manually, interactively by running `yarn run develop` to open and serve `development.html` with the current build of thebe.
+You can test manually, interactively by running `yarn run develop` to open and serve `development.html` with the current build of `thebe`.
+This allows you to develop and test something that you are implementing locally, for example a version of your page where you intend to use the library.
+
+When modifying the `thebe` library manually testing is still required as automated test coverage is limited. Changes to `thebe` can tested for regression in two ways:
+
+1.  via the `html` examples in the `examples/` folder
```
choldgraf, line 136:
```suggestion
1.  via the `html` examples in the `docs/_static/html_examples/` folder
```

PR #482, opened 2021-09-20
temporarily disable dependabot

stevejpurves:
Proposal to temporarily disable dependabot and consider dependency upgrades in one go ahead of releases in a more structured way.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,3 +7,4 @@ updates:
     directory: "/" # Location of yarn.lock
     schedule:
       interval: "weekly"
+    open-pull-requests-limit: 0


stevejpurves:
@moorepants @choldgraf suggest we do this until we're ready for next release as @moorepants suggested

moorepants:
Fine with me.

choldgraf:
Is the plan to add it back in? If so then we should track this somewhere so we don't forget.

Maybe we can create an issue that gives at date where we should turn this back on or something? Or at the least add a comment in the code

In general I'm +1 on it, i trust y'all's judgment here

stevejpurves:
I have opened #484 to track @choldgraf

PR #528, opened 2021-11-19
📚 [DOC] adding jsdocs to thebe API

stevejpurves:
We have had `jsdoc` in the docs build for a while but with no `jsdocstrings` in the code. 

This PR aims to add those for the main elements of the API.

One aspect of this that maybe needs discussion is around the options object and whether we generate documentation using `jsdoc` for that rather than having separate references in the docs and readme that is not coupled with the code. Part of that means changing the `options` from a javascript literal to a class that jsdoc can handle. This is potentially a good step in any case as it allows us to centralise validation and gives us a place to apply a schema, however we need to be careful that the interface for simple usage doesn't change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## 0.8.3 - To be confirmed
+
+### Deprecations
+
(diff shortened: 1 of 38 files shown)


stevejpurves:
I have a problem with my fork being our of sync with the main repo 👎🏼  will address that

PR #529, opened 2021-11-19
💿  [DOC] adding jsdoc generated API documentation

stevejpurves:
We have had `jsdoc` in the docs build for a while but with no `jsdocstrings` in the code.

This PR aims to add those for the main elements of the API.

One aspect of this that maybe needs discussion is around the options object and whether we generate documentation using `jsdoc` for that rather than having separate references in the docs and readme that is not coupled with the code. Part of that means changing the options from a javascript literal to a class that `jsdoc` can handle. This is potentially a good step in any case as it allows us to centralise validation and gives us a place to apply a schema, however we need to be careful that the interface for simple usage doesn't change.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -1,13 +1,32 @@
 =========
-thebe API
+Thebe API
 =========
 
-The thebe JavaScript API
+When adding Thebe to a page one only needs to know about the bootstrap process and option object.
+By configuring the options you can already manipulate
+
+.. js:autofunction:: thebelab.bootstrap
+
+Configuration Options
+---------------------
+
+.. js:autoclass:: Options
+    :members:
+    :exclude-members: Options
 
 .. note::
 
-    This document is a work in progress.
-    We need to add jsdoc-style docstrings to our exported functions.
+    Above we document the javascript Thebe Options class which provides the default options for Thebe
+    and is a good reference for all current options.
 
+    In typical usage, we add any desired options as an object literal in a script tag on our page,
+    see :doc:`./configure`. for details. Option specified on page will be override the defaults
+    listed above.
 
-.. js:autofunction:: thebe.bootstrap
+High Level JQuery API
+---------------------
+
+.. js:autofunction:: ./render.renderCell
+.. js:autofunction:: ./render.renderAllCells
+.. js:autofunction:: thebelab.mountActivateWidget
+.. js:autofunction:: thebelab.mountStatusWidget
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -5,7 +5,7 @@ Thebe API
 When adding Thebe to a page one only needs to know about the bootstrap process and option object.
 By configuring the options you can already manipulate
 
-.. js:autofunction:: thebelab.bootstrap
+.. js:autofunction:: thebe.bootstrap
 
 Configuration Options
 ---------------------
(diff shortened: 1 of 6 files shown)


docs/api.rst
```diff
@@ -1,13 +1,32 @@
...
+    listed above.
 
-.. js:autofunction:: thebe.bootstrap
+High Level JQuery API
+---------------------
+
+.. js:autofunction:: ./render.renderCell
+.. js:autofunction:: ./render.renderAllCells
+.. js:autofunction:: thebe.mountActivateWidget
+.. js:autofunction:: thebe.mountStatusWidget
```
stevejpurves, line 32: this is pretty straightforward addition of jsdoc to generate content for the main js functions

src/options.js
```diff
@@ -20,38 +20,205 @@ if (typeof window !== "undefined" && typeof window.define !== "undefined") {
   window.define("@jupyter-widgets/output", output);
 }
 
+/**
+ * Thebe options object
+ */
+export class Options {
```
stevejpurves, line 26: In order to document the options object in this way using jsdoc, we need to move away from the object literal and create an object class. This is potentially a good step as this is where we could also do some validation against a schema for the options. 

However, it also means that we are documenting options in. a few different places now, including in the main readme. It feels good to go with jsdoc generated docs for the options, as then the docs are generated from code as it is added / changed, so perhaps use this as the way to document the options and update the readme etc.. to point here.

choldgraf: As a general rule, I am fine with having a single source of truth for documentation and then simply referring to that. If our README is getting complex enough, we should cut it down to minimal information needed for people to get started, and then push the full docs into our website via jsdoc.

docs/api.rst
```diff
@@ -1,13 +1,32 @@
 =========
-thebe API
+Thebe API
 =========
 
-The thebe JavaScript API
+When adding Thebe to a page one only needs to know about the bootstrap process and option object.
+By configuring the options you can already manipulate
```
choldgraf, line 6: sentence seems like it's dangling?

docs/api.rst
```diff
@@ -1,13 +1,32 @@
...
 
 .. note::
 
-    This document is a work in progress.
-    We need to add jsdoc-style docstrings to our exported functions.
+    Above we document the javascript Thebe Options class which provides the default options for Thebe
+    and is a good reference for all current options.
 
+    In typical usage, we add any desired options as an object literal in a script tag on our page,
+    see :doc:`./configure`. for details. Option specified on page will be override the defaults
```
choldgraf, line 23:
```suggestion
    see :doc:`./configure`. for details. Options specified on page will be override the defaults
```

docs/api.rst
```diff
@@ -1,13 +1,32 @@
...
+    listed above.
 
-.. js:autofunction:: thebe.bootstrap
+High Level JQuery API
+---------------------
+
```
choldgraf, line 28: Could you start each section with a one-sentence description of what the section will contain?

docs/api.rst
```diff
@@ -1,13 +1,32 @@
 =========
-thebe API
+Thebe API
 =========
```
choldgraf, line 3: How about we call this `Javascript API` instead?

choldgraf (commented):
This looks really nice - I like that we're automatically generating these docs.

I think we now have some redundant information with https://thebe--529.org.readthedocs.build/en/529/config_reference.html as well - we should just have a single place with this information and cross-link to it.

stevejpurves:
Closing this as we need to move this to typedoc, so a different configuration is required and we need to now look at documenting thebe and thebe-core. This will either happen in the PR #554 or after that comes in

Issue #560, opened 2022-08-22
Cannot use Jupyterlite in `thebe` 0.9.0-rc0


stevejpurves:
### Describe the bug

**context**
When I do initialize thebe with the `useJuyterlite` option.

**expectation**
I expected `thebe` to initialize and start Jupyterlite server

**bug**
But instead we get an error in the JS console

![image (14)](https://user-images.githubusercontent.com/1473646/185902325-01cd2ff4-90dd-46b2-8781-f95780a316ed.png)

### Reproduce the bug

Add thebe in a script tag
set useJupyterlite: true in the config

stevejpurves:
Juptyerlite and all its dependencies load and work in `thebe-core` when bundled using webpack at that level. 

This bug in `thebe` which consumes `thebe-core` looks like `@lumino/commands` does not get bundled properly. Looking at it in the browser `devtools` and CommandRegistry is `undefined` although the import does succeed.

stevejpurves:
Changes to the bundling have been made which should get over the error reported above. Use of JuptyerLite is still blocked by this issue https://github.com/executablebooks/thebe/issues/575

PR #565, opened 2022-08-24
progress by winding back shiming

stevejpurves:
Build is still unstable but the error shown in #560 is no longer present.

I can reach this state after a clean rebuild but have hit other weird errors too.

The current best state it that all code is now loading, server and session initializes but we have a problem loading a wheel 

image

This PR also changes some of the demo build setup, and sets the `publicPath` to `auto` getting around previous chunk loading issues.


stevejpurves:
Winding back the shimming means package size has increased but the shining needs to be reassessed now that Juptyerlite is in place. There are potentially other ways to reduce package size, for example but using Jupyterlab's Codemirror class instead of loading and using a separate vanilla version of Codemirror. Some shimming can potentially be re-introduced too, but needs to be done in a very methodical way.

PR #568, opened 2022-10-18
feat/better runtime objects

stevejpurves:
This PR adds the following:

* overhaul of the runtime object making them more usable from web app contexts
* less dependence on messaging and a stronger promise based API
* changes to the ThebeManager and moving to one manager per notebook rather than per cell, improving and probably fixes various existing `ipywidgets` error cases
* added a react based demo app

This PR is the result of significant debugging into `ipywidgets` issues and attempted usage of `thebe-core` in a web app. bringing this in will help close off the thebe-code integration work, and allow us to generate an updated PR


stevejpurves:
failing tests here are issues with the test environment and the fact that we do not have a jupyter server running yet. Merging this PR and will deal with this on the `feat/integrate-thebe-core)` branch

PR #600, opened 2023-02-09
🪝hooks aligned for `ipynb` or plain code

stevejpurves:
The `react` package provides a couple of hooks to help with ThebeNotebooks. One of the hooks dealt wth `ipynb` files (`INotebookContent`) and had been developed further than the second hook that deals with plain code cells as an array of strings.

This PR aligns the base features of the hooks (various execute methods, common approach to ref handling, etc..) which still providing the two different loading methods.

note the original `useNotebookfromSource` has been renamed `useNotebookfromSourceLegacy` and marked as deprecated for now, but users should transition to the new `useNotebookFromSource` (note fixed camel-casing).

### DIFF
--- a/packages/react/src/hooks/notebook.ts
+++ b/packages/react/src/hooks/notebook.ts
@@ -11,52 +11,13 @@ interface ExecuteOptions {
   preprocessor?: (s: string) => string;
 }
 
-export function useNotebook(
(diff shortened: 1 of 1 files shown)


stevejpurves:
cc @rowanc1

Issue #606, opened 2023-02-21
'thebe-lite' upgrade to use pyolite `beta.18`


stevejpurves:
### Context

Versions of pyolite and associated packages >beta.14 - include an alternate kernel startup sequence that addresses #580 as well as bringing in latest improvements that we want to keep up to date with.

### Proposal

Update `thebe-lite` to use the latest `beta.18` release

### Tasks and updates

- [x] update thebe lite
- [x] everything builds
- [x] all tests pass
- [x] simple demo app - jupyerlite example works
- [x] simple demo app - jupyerlite ipywidgets example works
- [x] core demo app - jupyerlite example works
- [x] core demo app - jupyerlite ipywidgets example works
- ⛔️ deferred startup expects in #580 is seen

jtpio:
FYI @stevejpurves we are currently migrating the Pyodide kernel for JupyterLite to a separate repo: https://github.com/jupyterlite/jupyterlite/pull/854

Releases of the kernel will be made from https://github.com/jupyterlite/pyodide-kernel and released as `jupyterlite-pyodide-kernel` on PyPI. The wheel includes the prebuilt assets like the extra other wheels for starting the kernel, which means this should simplify how the kernel is used elsewhere like in `thebe` and should hopefully not require a `jupyter lite build`.

A first early release is available here: https://github.com/jupyterlite/pyodide-kernel/releases/tag/v0.0.2, in case you would like to try it out. There might still be some issues of course!

stevejpurves:
@jtpio thanks for the heads up - this is exactly what we were thinking about here, as trying to bundle the wheels and assets complicates usage for sure. I will try it out and let you know.

jtpio:
Another update that a new release of JupyterLite is out, which moved the core functionalities to a new `jupyterlite-core` package: https://github.com/jupyterlite/jupyterlite/releases/tag/v0.1.0b19

And the Pyodide kernel now depends on it: https://github.com/jupyterlite/pyodide-kernel/releases/tag/v0.0.5

Let us know if you find any issues, thanks!

stevejpurves:
closing this issue although #580 could still be improved

PR #607, opened 2023-02-21
📦 upgrade to juptyerlite **latest**

stevejpurves:
Work to address #606


stevejpurves:
ok, so `ipywidgets` are not working in the demo apps and we are not seeing the thebe status message come "after" the kernel is fully loaded -- which we expected to happen since `beta.15` -- needs some investigation

Issue #608, opened 2023-03-09
`interact` based `ipympl` plots not working as expected


stevejpurves:
### Describe the bug

**context**
When thebe is used to render an `ipympl` plot that has been created using interact, this fails to render as expected. Instead of getting a canvas based render of the output, it seems ot fallback to html.

When manually rendering in an output widget, thebe performs as expected.

The code to produce the plot. is:

```python
%matplotlib widget
import ipywidgets as widgets
import matplotlib.pyplot as plt
import numpy as np      
from ipywidgets import Output, FloatSlider, Layout
from IPython.display import clear_output

def sine_func(x, w, amp):
  return amp*np.sin(w*x)

fig_one, ax_one = plt.subplots(1,1)

x = np.linspace(0,10)

@widgets.interact(w=(0, 4, 0.25), amp=(0, 4, .1))
def update(w = 1, amp = 1):
    plt.figure(fig_one.number)
    plt.clf()
    plt.ylim(-4, 4)
    plt.plot(x, sine_func(x, w, amp))
    plt.show()
```

Which produces a `text/html` based plot on in the `OutputArea`

![image](https://user-images.githubusercontent.com/1473646/223967717-d6b99128-18f9-4d4c-a362-6dd400c1c56b.png)

**expectation**
I expected to be able to get the same behaviour from `interact`, `interactive_output` and manual use of an Output widget. The same notebook displays as expected in JupyterLab

**bug**
But instead I get a static, non interactive plot.

**problem**
This is a problem for people doing trying to create interactive figures, as the standard `ipympl` controls are not available and the plot will not respond to other click and mouse events.

### Reproduce the bug

Coming soon .. a link to the notebook

stevejpurves:
It turns out that removing the customized Output widget that was included when switching to the KernelWidgetManager, fixes this particular issue  https://github.com/executablebooks/thebe/blob/main/packages/core/src/output.ts

However it also (re)introduces errors when the widget manager is created use of the customized Output was based on similar changes in `voila` to remove those errors, seems like this is a side effect of those changes - so this could also be affecting voila? (cc @martinRenou)

Investigating to see if a different change to Output could be made to fix both problems

stevejpurves:
After debugging it's clear that the manager class nor the customized `Output` widget class are **not** causing the issue.

After more debugging and another set of eyes from @agoose77 👀 -- removing the `plt.show()` at the end of the `def update():` function gets this working in `thebe` and `jupyterlab`. 

[...]

PR #609, opened 2023-03-14
fix `ipympl` render issues

stevejpurves:
These changes now align behaviour for `ipympl` based plots and certain widget configurations between `thebe`, `voila` and `jupyterlab`.

Specifically addresses: #608


stevejpurves:
As a part of this PR a minimal demo of the `thebe-react` functionality has been added

PR #612, opened 2023-04-04
use latest Jupyterlite with programatic options

stevejpurves:
The purpose of this PR is to update to using the latest `juptyerlite` codebase, this involves adjusting to the new organisation of the `@juptyerlite/pyodide-kernel` and taking steps to specify the location of the wheels to load. This should be a big improvement over the previous implementation once in place, and the objective is to be able to load the latest wheels from CDN (possibly direct from unpkg.com).

### DIFF
--- a/apps/simple/static/lite.html
+++ b/apps/simple/static/lite.html
@@ -11,14 +11,28 @@
         useBinder: false
       }
     </script>
-    <script id="jupyter-config-data" type="application/json">
-      {}
-    </script>
-    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/4.7.0/css/font-awesome.min.css" crossorigin="anonymous" referrerpolicy="no-referrer" />
+    <!-- <script id="jupyter-config-data" type="application/json">
+      {
+        "litePluginSettings": {
+          "@jupyterlite/pyodide-kernel-extension:kernel": {
+            "pipliteUrls": ["https://unpkg.com/@jupyterlite/pyodide-kernel@0.0.6/pypi/all.json"]
+          }
+        }
+      }
+    </script> -->
+    <link
+      rel="stylesheet"
+      href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/4.7.0/css/font-awesome.min.css"
+      crossorigin="anonymous"
+      referrerpolicy="no-referrer"
+    />
     <link rel="stylesheet" href="thebe.css" />
     <script src="thebe-lite.min.js"></script>
     <link rel="stylesheet" href="main.css" />
-    <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/jupyter-matplotlib@0.11.2/css/mpl_widget.css" />
+    <link
+      rel="stylesheet"
+      href="https://cdn.jsdelivr.net/npm/jupyter-matplotlib@0.11.2/css/mpl_widget.css"
+    />
     <script src="index.js"></script>
   </head>
   <body>
--- a/packages/lite/src/jlite.ts
+++ b/packages/lite/src/jlite.ts
@@ -33,8 +33,17 @@ function* activePlugins(extension: any) {
   }
 }
 
+/**
+ * Example litePluginSettings shape
+ * 
+ * '@jupyterlite/pyodide-kernel-extension:kernel': {
+        pipliteUrls: ['https://unpkg.com/@jupyterlite/pyodide-kernel@0.0.6/pypi/all.json'],
+        piplistWheelUrl: 'https://some-url.com'
+      }
+ * 
+ */
+
 type LiteServerConfig = {
-  federatedExtensions: Record<string, any>;
   litePluginSettings: Record<string, any>;
 };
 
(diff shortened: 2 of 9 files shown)


apps/simple/static/lite.html
```diff
@@ -11,14 +11,28 @@
...
-    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/4.7.0/css/font-awesome.min.css" crossorigin="anonymous" referrerpolicy="no-referrer" />
+    <!-- <script id="jupyter-config-data" type="application/json">
+      {
+        "litePluginSettings": {
+          "@jupyterlite/pyodide-kernel-extension:kernel": {
+            "pipliteUrls": ["https://unpkg.com/@jupyterlite/pyodide-kernel@0.0.6/pypi/all.json"]
+          }
+        }
+      }
+    </script> -->
```
stevejpurves, line 22: This option could / should work as er make a call to `PageConfig.getOptions('')` on starting the server?

packages/lite/src/jlite.ts
```diff
@@ -1,32 +1,76 @@
...
+   * and set options on the page config directly
+   */
+  PageConfig.setOption(
+    'litePluginSettings',
+    JSON.stringify({
+      '@jupyterlite/pyodide-kernel-extension:kernel': {
+        pipliteUrls: ['https://unpkg.com/@jupyterlite/pyodide-kernel@0.0.6/pypi/all.json'],
+      },
+    }),
+  );
```
stevejpurves, line 58: we should be able to set the `litePluginSettings` here? and this would be ahead of the `@jupyterlite/pyodide-kernel-extension` being activated, so it should be able to pick up on these?

packages/lite/src/jlite.ts
```diff
@@ -1,32 +1,76 @@
...
+   *
+   * liteExtension - essential, as these are how kernels are added
+   * federatedExtension - general jupyterlab extensions
+   * federatedMimeExtension - render extensions? e.g. @jupyterlab/javascript-extension, @jupyterlab/json-extension
+   * federatedStyles - ?
+   *
+   * TODO we're not suppporting all of these yet
+   */
+
+  console.log('getOption', JSON.parse(PageConfig.getOption('litePluginSettings')));
```
stevejpurves, line 71: this confirms in that the options are set as expected

stevejpurves:
currently when trying to run `jlite.html` in the simple demo, we hit this problem during kernel initialization:

image

I was expecting that setting the `litePluginOptions` appropriately would address this but we're not seeing even a network call being made to the `all.json`.

jtpio:
The screenshot suggests the name is malformed, but they seem to look good in the npm package: https://www.npmjs.com/package/@jupyterlite/pyodide-kernel?activeTab=code

image

If you use the `@jupyterlite/pyodide-kernel` directly then it should be possible to import them from `/pypi` yes.

Maybe we could also check the Web Worker is correctly getting the wheel info.

stevejpurves:
@jtpio yes the wheels look good in the package - what I am seeing though is webpack is adding them to my distribution and replacing the names with hashes, so my dist folder contains randomly hashed `.whl` files.

jtpio:
> so my dist folder contains randomly hashed `.whl` files.

You might need to use the Asset Modules with a `generator` like here: https://github.com/jupyterlite/pyodide-kernel/blob/01ef2f5ad3c468205d78dddeffa2990c0bd4eee9/packages/pyodide-kernel-extension/webpack.config.js#L8-L12

apps/simple/static/lite.html
jtpio: Right the kernel should be able to pick it up here: https://github.com/jupyterlite/pyodide-kernel/blob/01ef2f5ad3c468205d78dddeffa2990c0bd4eee9/packages/pyodide-kernel-extension/src/index.ts#L45

packages/lite/src/jlite.ts
jtpio: > so it should be able to pick up on these?

Yes I would assume so.

stevejpurves: cool - and yes I am seeing some impact of setting this but it's not loading the `piplite` wheel from `unpkg` for some reason.

i.e. **with** setting the `litePluginSettings` here, on initializing the kernel two of the wheels are loaded from unpkg.com 👏 **but** it still looks to the `/pypi/all.json` locally and still loads `piplite` from there 🤔

image

stevejpurves: ok so this kind of makes sense as `litePluginSettings` is setting the `pipliteUrls` i.e. where piplite will load it's wheels from -- we need to also configure something separate to direct where the `piplite` wheel itself will be loaded from...?

stevejpurves: opened https://github.com/jupyterlite/pyodide-kernel/pull/42 to address this

stevejpurves: I have also started looking at the remote federated module loading, which I think I'll start in a separate PR on top of this one.

I'm just wondering though, if we put that remote loading in place it will immediately supercede this arrangement of build-time including the server extension while picking up the wheels from online - so it's just better to move to that now. Then AFAIK this will mean that the kernel can be specified by option at runtime, and also be interchangeable (i.e. we can load other xeus kernels too)

stevejpurves:
ok I have changed the pageconfig to set the `litePluginOptions` based on an optional argument in the start server function. As it is this would allow runtime changes to point piplite to look for wheels at a remote location. This means that the iolie wheel itself still needs to be hosted locally but if https://github.com/jupyterlite/pyodide-kernel/pull/42 is merged and released then all wheels can be configured remotely

stevejpurves:
merged https://github.com/executablebooks/thebe/pull/612 which addresses a lot of this but we still need some testing to see if that last todo on startup signalling has been acheived.

Issue #460, opened 2021-08-16
Can't change theme to cobalt


FasihZafar:
### Describe the problem

I have followed the contribution guide and managed to get there up and running on local host. After which I tried switching theme value to cobalt inside the script tag:
    <script type="text/x-thebe-config">
      {
        requestKernel: true,
        binderOptions: {
          repo: "binder-examples/requirements",
        },
        codeMirrorConfig: {
          "theme": "cobalt"
        },
      }
    </script>
But nothing changes.

### Steps to reproduce

1. follow the instructions from here https://thebe.readthedocs.io/en/latest/contribute.html#set-up-a-development-environment
2. add the them key value pair in the codeMirrorConfig dictionary

stevejpurves:
Hi @FasihZafar 

Thanks for reporting. So it looks like the codemirror theme is not being applied at all, and I don't think there is code in place to actually implement it!

In order to apply codemirror theme, it is necessary to load the appropriate `css` file which thebe is not doing for you. However, even if I manually add the `cobalt.css` to my page, and set the option it still does not work. Perhaps we need configure codemirror differently in the version we on which is `5.61.1`.

Definitely an issue though.

Issue #256, opened 2020-10-17
Improve styling of thebe elements


akhmerov:
After migration to the sphinx-book-theme, thebe elements look like this upon activation:

![image](https://user-images.githubusercontent.com/2069677/96334936-be4e8600-1074-11eb-9e49-017315da4f11.png)

I think this would look better without line separators in the source code, and the button styling could be improved.

akhmerov:
Also the code really doesn't look like CodeMirror abcdef theme...

moorepants:
#184 is relevant. Maybe it needs to ship with minimal styling.

Issue #241, opened 2020-10-14
Codemirror does not take effect


linjie98:
Hi @minrk I tried to add a theme to the editor through the following configuration, but it didn't work.
I read the previous issues, but it still doesn’t work
`<script type="text/x-thebe-config">
    {
      bootstrap: true,
      kernelOptions: {
        name: "python3",
        serverSettings: {
          "baseUrl": "http://127.0.0.1:8888",
          "token": "test-secret"
        }
      },
      codeMirrorConfig: {
          theme: "abcdef"
      },
    }
  </script>`
Can you tell me what should I do, thanks

ymdahi:
Just ran into this issue. Would be nice to be able to add some style/have control over the code editor, like https://codemirror.net/demo/theme.html#eclipse

stevejpurves:
this is now functional see https://executablebooks.github.io/thebe/ for `darcula` in place

Issue #223, opened 2020-06-25
Move codeMirrorOptions outside binderOptions?


choldgraf:
I noticed that right now thebelab is looking inside `binderOptions` for the codeMirror config:

https://github.com/minrk/thebelab/blob/8b6d82c24727ae67685c89bea99f38031df12ea1/src/thebelab.js#L263

Why is that? Isn't CM something that would be used regardless of the backend supplying the kernel?

stevejpurves:
done! see now functional see https://executablebooks.github.io/thebe/
Some draft docs in `apps/docs-core`

PR #622, opened 2023-04-27
📡 autoload mode and strip back bundled css classes and overrides

stevejpurves:
This PR is work towards fixing syntax highlighting, mode selection and theming which will impact on the following issues:

* [x] https://github.com/executablebooks/thebe/issues/460
* [x] https://github.com/executablebooks/thebe/issues/241
* [x] https://github.com/executablebooks/thebe/issues/256

Useful reference for theme changes: https://codemirror.net/5/demo/theme.html#darcula

We should also tackle this here: 
* [x] https://github.com/executablebooks/thebe/issues/223 
it's in theory a breaking change to API but no-one can be relying on these options anyway.

### DIFF
--- a/apps/docs-core/thebe-docs.md
+++ b/apps/docs-core/thebe-docs.md
@@ -1,0 +1,34 @@
+# Plceholder for thebe related docs
+
+## Codemirror setup
+
+The following is available in the thebe options object:
+
+```typescript
+{
+  codeMirrorConfig?: {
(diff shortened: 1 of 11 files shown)


stevejpurves:
work so far - codemirror now looks active! and should react properly to the language in each cell even, but there are still styling issues:

image

stevejpurves:
css issues fixed, options extended and theme support working provided you load additional css where needed:

image

Issue #630, opened 2023-05-30
Change `thebe-core` object interfaces to force external `RenderMimeRegistry` creation


stevejpurves:
When using `thebe-core` and `thebe-react` one of the major sources of issues with `ipywidgets` in notebooks is related to the creation and management of the `RenderMimeRegistry`. 

At the moment both the `ThebeSession` and `ThebeNotebook` are quite permissive and will allow you to create either without supplying a `rendermime` argument, that argument is optional and when `undefined` will cause the creation of a rendermime registry in each object.

Having two rendermime registries floating around is a source of error, seems to lead to timing related issues in both `thebe-core` and `thebe-react`. It's a non-issue in `thebe` as the usage of the session and notebook objects there is quite fixed, with a strict 1-to-1 notebook & session relationship that is hardcoded into the "attach" sequence.

In other apps though, the developer has flexibility of how notebooks and sessions are related, and in which order they may be created and made ready. 

This can lead to subtle and hard to debug issues, often that lead to having to figure our if there are multiple rendermime instances in play and what led to their independent creation -- this is hard to debug! 🤯

We need to make some changes to improve this situation asap.

1. **Is the ThebeManager in the right place?** -- The `ThebeSession` holds the instance of the `ThebeManager` we need to manage `ipywidgets` models, and while notebooks could hold one each (enforcing a widget boundary at the notebook level) it's more natural to have a single widget manager for the session, as its' model listing will reflect the models available on the session/kernel. So it seems like this is ok, and the manager should stay here with the session/kernel.

2. **Should the ThebeSession be the object to hold the rendermime?** -- this would mean if notebooks were created first, we would have to completely refresh all cells' output areas when they attach to a session. this would mean swapping out any local or temporary rendermime instances for the instance within the session | widgetmanager, and probably recreating the cells and their output areas. This is could be done, may mean reconnecting OutputArea widgets to the DOM which could produce some effects. It's also a silent behaviour, so add a little more magic for the developer.

3. **Should we make creation of a RenderMimeRegistry an explicit step for the `thebe-core` and `thebe-react` developer?** -- this would mean we could remove the optional parameter and enforce that a single rendermime is passed in everywhere it is needed. This would lead to less magic but gives the developer another thing to think about, this is not so bad a understanding rendermime is key to `ipywidgets` working well and forces the developer to manage how rendermimes are distributed in the app. 

[...]

PR #632, opened 2023-05-31
🪴external rendermime managment

stevejpurves:
in #630 I was discussing how to improve the current very buggy rendermime creation. I pushed though on option (3) [making rendermime registry creation an explicit part of the API] in this PR, and...

Things are immediately much better, specific improvements are:

* much easier to reason about the expected ipywidgets behaviour
* some confusing addWidgetRenderer calls in notebooks have been removed
* the ThebeManager implementation is now just simple without the bizzare behaviour of adding widgets to some other rendermime.
* the only place where a rendermime registry is completed is now in the `PassiveCellRenderer` which feels right as that class is intended for standalone usage on static content.

So this is a keeper 🤘

### DIFF
--- a/apps/demo-core/src/app.ts
+++ b/apps/demo-core/src/app.ts
@@ -1,5 +1,6 @@
 import type { ThebeSession } from 'thebe-core';
 import {
+  makeRenderMimeRegistry,
   ThebeEvents,
   ThebeEventType,
   shortId,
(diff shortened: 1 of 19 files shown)


rowanc1:
This is awesome to see, really like the simplifications and explicit passing of the rendermime!

Issue #638, opened 2023-06-18
Error when running demo-react 


shaielc:
### Describe the bug

**context**
Cloning and trying to run `demor-react`

**bug**
got an error `Uncaught SyntaxError: Unexpected token '<'` in `thebe-lite.min.js:1 `

### Reproduce the bug

1. clone repo
2. `cd apps/demo-react`
3. `mkdir node_modules`
4. `npm i`
5. `npm start`

stevejpurves:
hi @shaielc - the demos are configured to be build and run from the builds within the monorepo, so I think the issue here is that you need to build the whole monorepo.

For your situation, what I would do now is:

1. delete the `node_modules` folder and any package-lock.json from `apps/demo-react`

This gets you back to the same as a clean clone.

1. cd into the root of the monorepo
2. `npm install`
3. `npm run build`
4. `npm run start:react`

After you have done this once, then starting the demo from the `apps/demo-react` folder will probably work as expected.

stevejpurves:
also I can see the ourve of the error in your console.

[this line](https://github.com/executablebooks/thebe/blob/3d0d729a9257686d9c364fcb6b13ac01ad0c0a85/apps/demo-react/public/index.html#L6) should be removed as it is now redundant. I'm about to commit a change that will remove that.

shaielc:
I am not sure about #642 after pulling #641 I can run `npm i && npm start` in `apps/demo-react` and it works fine.

stevejpurves:
ok - glad to hear it! I'm usually building the whole monorepo and starting from the root but glad to hear this is working for you

PR #642, opened 2023-06-20
doc: update setup

shaielc:
Updated `react-demo` README following issue #638.

### DIFF
--- a/apps/demo-react/README.md
+++ b/apps/demo-react/README.md
@@ -1,7 +1,15 @@
-# Getting Started with Create React App
+# Setup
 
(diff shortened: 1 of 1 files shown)


stevejpurves:
this is great thanks @shaielc -- that boilerplate readme wasn;t useful before.

In light of this I've opened #643 as improving the docs for the other demos and surfacing all of that in the main docs site is a good thing to have

stevejpurves:
bringing this in as apparently the preview will not work from a fork until the next version of the action we are using, hence the failure

PR #646, opened 2023-06-23
docs update : Installation copy-assets script

### LINKED ISSUE #645: Add `copy-thebe-assets.cjs` script to the package distro
As discussed here https://github.com/executablebooks/thebe/issues/637 we should ship the `copy-thebe-assets.cjs` so it can be used directly by consumers.

shaielc:
Added the `copy-thebe-assets` script and referenced it in the relevant docs.

This resolves #645 

If I am understanding correctly this stage is required whenever using either `thebe-lite` or `thebe-core`,  so I put the script in `thebe-core`.

I have also made it optional to have `thebe-lite` installed but later I saw the it is a dependency of `thebe-core` which makes this change kinda moot

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -107,5 +107,8 @@
   },
   "nohoist": [
     "requirejs"
-  ]
+  ],
+  "bin": {
+    "copy-thebe-assets": "./bin/copy-thebe-assets.cjs"
+  }
 }
--- a/apps/docs-core/react-lite.md
+++ b/apps/docs-core/react-lite.md
@@ -27,4 +27,7 @@ In the example above with `publicPath="/thebe"`, the script tags would be pointi
 Your build system should be updated to copy the bundles into place appropraitely.
 
 **Note:** currently `thebe-react` is configured to always load the `pyodide` and other base wheels from CDN, so there is no need to deploy those.
+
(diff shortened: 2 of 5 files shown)


shaielc:
I have tested it by running:
```
cd packages/core
npm run build
npm pack
mkdir /tmp/test
cd /tmp/test
npm init
npm install <path to rep>/packages/core/thebe-core-0.2.3.tgz
npx copy-thebe-assets thebe
```

packages/core/package.json
```diff
@@ -107,5 +107,8 @@
   },
   "nohoist": [
     "requirejs"
-  ]
+  ],
+  "bin": {
+    "copy-thebe-assets": "./bin/copy-thebe-assets.cjs"
```
rowanc1, line 112: This is great. I am using this over in another myst-theme as well, and it will be nice to not have to copy this script all over the place!

shaielc:
The resource not accessible issue seems like a problen with the CI 
I am guessing the cause is that this PR is from a fork

stevejpurves:
@shaielc yes that's a limitation with the preview action.
Thanks for your PR!

stevejpurves (approved):

stevejpurves:
`v0.2.4` of the packages have been released which contain this change!

PR #650, opened 2023-07-03
fix/timeouts

stevejpurves:
async loading can fail on slower networks - default timeout and number of attempts has been increased as well as optional parameters added.


rowanc1 (approved):
🚀

PR #685, opened 2023-09-05
fix/server blatting

stevejpurves:
The `ThebeServerProvider` could replace active servers during navigaion easily, these changes protect against that by checking to see if the server is already "ready" and has a user server url.

This complements well the existing `disconnect` functionality.

### DIFF
--- a/.changeset/metal-humans-beg.md
+++ b/.changeset/metal-humans-beg.md
@@ -1,0 +1,5 @@
+---
+'thebe-core': patch
+---
+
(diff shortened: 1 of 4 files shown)


stevejpurves:
A better check on the user server should include pinging the `/api/status` endpoint. This would be asynchronous so would mean changing the main side effect in the provider to be an explicit action via the `connect` function, this seems more sensible than the current effect cascade - needs some testing out to understand the impact on consuming applications.

PR #693, opened 2023-10-18
typos

psychemedia:
I ran a spellchecker over some of the docs and fixed the obvious typos.

There's some inconsistency in capitalisation around Jupyter/jupyter etc, and some variations in en-us vs en-gb spellings, but I'm not sure what style you prefer or whether you're keen to enforce it. Plus that sort of copy-editing can take forever!;-)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,7 +6,7 @@
 
 The main `thebe` packages are located in `packages/` and are:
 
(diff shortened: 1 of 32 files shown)


stevejpurves:
thnks @psychemedia - we're aiming for `en-us` but because I'm here, we're ending up with some `en-gb` as I can't shake good habits.

Issue #692, opened 2023-10-18
Using the current dev version with Jupyter Books


psychemedia:
### Context

I'm hoping to return to using Jupyter Book and `thebe` to do some interactive learning materials demo with a JupyterLite backand and was wondering what the simplest route to getting started for non-developers is.

IIRC, the process used to be `pip install jupyter-book thebe`, a simple `_config.yml` and `_toc.yml` file, a markdown file with a `jupytext` header, and a `{code-cell}` admonition block with the code intended for execution, and a simple `thebe` setting in the confg file to use MyBinder. But I'm not sure how to get started now?

The PyPi package is the legacy (pre-JupyterLite) version, and this repo does not appear to something that is intended to be installed as `pip install https://github.com/executablebooks/thebe/archive/refs/heads/main.zip`.

I note there is a [`sphinx-thebe`](https://sphinx-thebe.readthedocs.io/en/latest/index.html) package, but it also looks to be a legacy package without JupyterLite backend support.

### Proposal

Dead simple instructions for building a Jupyter Book with `thebe` JupyterLite support.

I'm happy to try to help put docs together around a set of bare bones steps, and maybe even have a go at the `sphinx-thebe` extension to make things easier for non-devs, but at the moment I'm not sure what's a pragmatic way to get started.

psychemedia:
I wonder.. would the simplest way be a `sphinx-theme` that extends `sphinx_book_theme` and loads in the additional assets?

stevejpurves:
Hi @psychemedia!

You're right in that thebe is not a python package, and can't be installed from pypi.

The integration of the 0.9.0-rc.X branch back into jbook untested so far, I've been focussing on the integration into `mystmd` which is a few steps ahead. 

But I believe getting `sphinx-thebe` to use the latest `0.9.0-rc.9` version and testing from there is the way forward. API and behaviour should be the same as the `0.8.x` branch but it really needs tested before anything could be released.

So I believe that one way forward would be to (give `mystmd` a try :) or) get your jbook up and running locally with thebe, and then probably make a local install of `sphinx-thebe` (git clone ..., then `pip install -e sphinx-thebe`), bump the THEBE_VERSION ([here](https://github.com/executablebooks/sphinx-thebe/blob/main/sphinx_thebe/__init__.py)) and test from there.

Regarding juptyerlite support though:

1. this probably needs changes in `sphinx-book-theme` to allow an additional thebe option through
[...]

Issue #689, opened 2023-09-28
thebe-lite demo not working due to missing ipywidgets wheel


stevejpurves:
### Describe the bug

**context**
When I do visit the online demo https://executablebooks.github.io/thebe/ and run it

**expectation**
I expected the code cell to produce an interactive output

**bug**
But instead an error is reported

```console
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[1], line 1
----> 1 await __import__("piplite").install(**{'requirements': ['ipywidgets']})
      3 import numpy as np
      4 import matplotlib.pyplot as plt

File /lib/python3.11/site-packages/piplite/piplite.py:103, in _install(requirements, keep_going, deps, credentials, pre)
    101 """Invoke micropip.install with a patch to get data from local indexes"""
    102 with patch("micropip._micropip._get_pypi_json", _get_pypi_json):
--> 103     return await _micropip.install(
    104         requirements=requirements,
    105         keep_going=keep_going,
    106         deps=deps,
    107         credentials=credentials,
    108         pre=pre,
    109     )

File /lib/python3.11/site-packages/micropip/_micropip.py:576, in install(requirements, keep_going, deps, credentials, pre)
    566 wheel_base = Path(getsitepackages()[0])
    568 transaction = Transaction(
    569     ctx=ctx,
    570     ctx_extras=[],
   (...)
    574     fetch_kwargs=fetch_kwargs,
    575 )
--> 576 await transaction.gather_requirements(requirements)
    578 if transaction.failed:
    579     failed_requirements = ", ".join([f"'{req}'" for req in transaction.failed])

File /lib/python3.11/site-packages/micropip/_micropip.py:342, in Transaction.gather_requirements(self, requirements)
    339 for requirement in requirements:
    340     requirement_promises.append(self.add_requirement(requirement))
--> 342 await gather(*requirement_promises)

File /lib/python3.11/site-packages/micropip/_micropip.py:349, in Transaction.add_requirement(self, req)
    346     return await self.add_requirement_inner(req)
    348 if not urlparse(req).path.endswith(".whl"):
--> 349     return await self.add_requirement_inner(Requirement(req))
    351 # custom download location
    352 wheel = WheelInfo.from_url(req)

File /lib/python3.11/site-packages/micropip/_micropip.py:457, in Transaction.add_requirement_inner(self, req)
    452 if self.check_version_satisfied(req):
    453     # Maybe while we were downloading pypi_json some other branch
[...]

Issue #688, opened 2023-09-28
Provide configuration options for use of the service worker for juptyerlite


stevejpurves:
Make use of recent changes allowing for ServiceWorker configuration in jupyterlite https://github.com/jupyterlite/jupyterlite/pull/1159#issuecomment-1731835761

PR #700, opened 2023-11-14
🚧 update jupyterlite and jupyterlab dependencies

stevejpurves:
This will bring all the jupyter deps up to date and enable us to address #688 #689 #692


stevejpurves:
The last commit dropped tests from the `thebe` subpackage, there have been getting more difficult to maintain as the module eco system has changed under us but now jest is just not building and switching out to `vitest` is not an easy thing because of some of our upstream deps.

have disabled the tests until we can overhaul this, we should focus testing efforts on the underlying `thebe-core` and then get a suitable `e2e` style tst going with new tooling (maybe cypress) for the `thebe` subpackage. This should perhaph happen after we transition to ESM modules + bundles as our build targets.

PR #710, opened 2023-11-21
🧑🏾‍🏭 override default jupyterlite service worker

stevejpurves:
added a reduced version of the jupyterlite service worker and build config to include it by default. This version should do no caching and only handle the braodcast messaging needed by jupyterlite.

addresses #688

### DIFF
--- a/.changeset/heavy-ravens-crash.md
+++ b/.changeset/heavy-ravens-crash.md
@@ -1,0 +1,5 @@
+---
+'thebe-lite': minor
+---
+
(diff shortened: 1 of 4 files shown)


rowanc1:
This is awesome. Nice work @stevejpurves!

jtpio:
Nice to see this can be overridden downstream :+1: 

Thinking we should eventually look into fixing that in JupyterLite either via config, or by moving the caching logic so it can be disabled more easily (as discussed in https://github.com/jupyterlite/jupyterlite/issues/1103).

PR #751, opened 2024-06-11
✊ update pyodide dependencies

### LINKED ISSUE #750: `thebe-lite` fails to run
### Describe the bug

**context**
thebe-lite error: cannot load the wheel for widgetnbextension

**expectation**
This is expected when there is a new widgetnbextension release

**solution**
Bump local dependencies

### Reproduce the bug

use the demo site

### List your environment

n/a

stevejpurves:
fixes #750


stevejpurves:
@jtpio  I've been looking back into the `pyodide-kernel` repo and see that `widgetsnbextensions` is effectively a no-op/mock there? is that correct, why is it needed? is it because some JupyterLab (or other) code a automatically tries to load that extension when `ipywidgets` are in play. Would love to understand it more.

jtpio:
iirc, it's because `widgetsnbextensions` when installing `ipywidgets` on the fly in the notebook, for example with `%pip install ipywidgets`. With the dependency on `widgetsnbextensions` being defined here: https://github.com/jupyter-widgets/ipywidgets/blob/bb2edf78e7dac26e4b15522a267d7b477026a840/python/ipywidgets/setup.cfg#L40

stevejpurves:
thanks @jtpio that makes sense!

PR #792, opened 2025-05-14
[doc] Update executablebooks to jupyter-book github org

choldgraf:
This updates a few links to point to `jupyter-book` instead of `executablebooks/`. I tried not to touch anything in the tests, since this is more for public docs / branding and I didn't want to chance breaking anything. I think that's fine.

I'll plan to merge this in a day or two if nobody objects!


choldgraf:
The CI seems broken, but given that this is just touching documentation, I am going to go ahead and self-merge this since it's just updating URLs to be the correct ones. I hope that's OK! I am happy to revert if this breaks something.

